PR #13, opened 2021-02-24
@W-8913847 - Fix test. Change message.

sfdctaka:


src/common/CommonUtils.ts
```diff
@@ -78,12 +78,22 @@ export class CommonUtils {
...
+        return mkdtemp(tempFolderPath)
             .then((folder) => {
                 return Promise.resolve(folder);
             })
             .catch((error) => {
-                return Promise.reject(error);
+                return Promise.reject(
+                    new SfdxError(
+                        util.format(
+                            'Could not create a temp folder at %s: %s',
```
sfdctaka, line 90: It's one line of formatting text but should I put it in a json file?

maliroteh-sf: technically it should be, but I'm willing to look the other way on this one mainly b/c (a) we have other instances of such things that we need to gather all into message file(s) and (b) I'd like to propose a change to how we organize strings in files (which would require further team discussion).

maliroteh-sf (approved):
👍

PR #15, opened 2021-02-24
@W-8913847 - Simplifying createTempDirectory.

sfdctaka:
`createTempDirectory` will no longer take user input for subfolder. Instead, it will use prefix that will be combined with random string of characters provided by `mkdtmp()`.

### DIFF
--- a/src/common/CommonUtils.ts
+++ b/src/common/CommonUtils.ts
@@ -4,10 +4,13 @@
  * SPDX-License-Identifier: MIT
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
-import { Logger } from '@salesforce/core';
+import { Logger, SfdxError } from '@salesforce/core';
 import * as childProcess from 'child_process';
 import { cli } from 'cli-ux';
 import fs from 'fs';
+import util from 'util';
+import path from 'path';
+import os from 'os';
 
 type StdioOptions = childProcess.StdioOptions;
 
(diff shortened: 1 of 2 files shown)


src/common/CommonUtils.ts
```diff
@@ -74,29 +71,6 @@ export class CommonUtils {
         });
     }
 
-    public static async createTempDirectory(
```
khawkins, line 77: As I stated offline, I think we should keep this. It's the right pattern for guaranteeing a unique working directory.

I think the only thing we really need to do here is kill the input parameter, and just use a literal for the folder prefix. It's our code, after all, and we can create the folder we want.  If we instead do:
```typescript
const mkdtemp = util.promisify(fs.mkdtemp);
const folderPrefix = 'lwc-mobile-';
const tempFolderPath = path.join(os.tmpdir(), folderPrefix);
return mkdtemp(tempFolderPath) // ...
```
We're going to get temp directories created like `/var/folders/ph/x5q98_c53kv31x739xcx2z5c0000gp/T/lwc-mobile-9jjMpW` on macOS, `/tmp/lwc-mobile-9jjMpW` on Linux, and `C:\Users\...\AppData\Local\Temp\lwc-mobile-9jjMpW` on Windows.

This is what we want. It's the right pattern for creating a disposable working directory. We should just fix what's broken in our code.

khawkins (approved):

maliroteh-sf (approved):

PR #18, opened 2021-02-25
Return API packages with emulator images

maliroteh-sf:
1. Updated `AndroidUtils.fetchAllAvailableApiPackages` to optionally only return those packages for which matching emulator images are installed.
2. Added comments to all public functions in `AndroidUtils`
3. Renamed `AndroidUtils.pollDeviceStatus` to `AndroidUtils.waitUntilDeviceIsReady` to match its iOS counterpart.
4. Renamed all instances of `find` and `required` in `AndroidUtils` function names to `fetch` and `supported` for consistency.
5. A few other renaming for the sake of consistency

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/lwc-dev-mobile-core",
-  "description": "Salesforce CLI plugin for mobile extensions to Local Development",
-  "version": "1.0.0-beta1",
+  "description": "Core module supporting Salesforce CLI mobile extension plug-ins",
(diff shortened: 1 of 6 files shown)


khawkins:
Do you mind throwing:
```json
"version": "1.0.0-beta2",
```
Into `package.json`, as part of your PR? We need to rev the version anyway, and the description incidentally needs updating.

khawkins (approved):
This should work better. It still has the artificial constraint of tying platform API to system image API, but at least it ensures that the requested image will be considered.

PR #19, opened 2021-03-01
change logic to detect when AVD is booted

maliroteh-sf:
1. Change logic to detect when AVD is booted
2. Introduce Promise that can be timed out
3. Remove the check for `lwc-dev-server` from base setup requirements
4. Fix issues with AVD config file that prevented an AVD to be launched from AVD Manager

### DIFF
--- a/src/common/AndroidUtils.ts
+++ b/src/common/AndroidUtils.ts
@@ -457,9 +457,8 @@ export class AndroidUtils {
                             )
                         );
                     });
-                    child.stderr.on('data', (data) => {
-                        const msg = data.toString() as string;
-                        if (msg && msg.includes('Error:')) {
+                    child.stderr.on('data', (data: Buffer) => {
+                        if (data.includes('Error:')) {
                             reject(
                                 new Error(
                                     `Could not create emulator. Command failed: ${createAvdCommand}\n${data}`
--- a/src/common/AndroidUtils.ts
+++ b/src/common/AndroidUtils.ts
@@ -458,11 +458,14 @@ export class AndroidUtils {
                         );
                     });
                     child.stderr.on('data', (data) => {
-                        reject(
-                            new Error(
-                                `Could not create emulator. Command failed: ${createAvdCommand}\n${data}`
-                            )
-                        );
+                        const msg = data.toString() as string;
+                        if (msg && msg.includes('Error:')) {
+                            reject(
+                                new Error(
+                                    `Could not create emulator. Command failed: ${createAvdCommand}\n${data}`
+                                )
+                            );
+                        }
                     });
                 } else {
                     reject(new Error(`Could not create emulator.`));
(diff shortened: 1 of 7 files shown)


src/common/AndroidUtils.ts
```diff
@@ -458,11 +458,14 @@ export class AndroidUtils {
...
-                            )
-                        );
+                        const msg = data.toString() as string;
+                        if (msg && msg.includes('Error:')) {
+                            reject(
+                                new Error(
+                                    `Could not create emulator. Command failed: ${createAvdCommand}\n${data}`
+                                )
+                            );
+                        }
```
maliroteh-sf, line 468: For some reason ADB also send non-errors (such as `Warning:`) to the `stderr` too 🤦  So we explicitly check for `Error:` to see if it is truly an error or not.

src/common/AndroidUtils.ts
```diff
@@ -533,33 +536,19 @@ export class AndroidUtils {
     public static async waitUntilDeviceIsReady(
         portNumber: number
     ): Promise<void> {
-        const command = `${AndroidUtils.getAdbShellCommand()} -s emulator-${portNumber} wait-for-device shell getprop sys.boot_completed`;
+        const quote = process.platform === WINDOWS_OS ? '"' : "'";
+        const bootCmd = `shell ${quote}while [[ -z $(getprop sys.boot_completed) ]]; do sleep 1; done;${quote}`;
```
maliroteh-sf, line 540: Read through this [SO post](https://stackoverflow.com/questions/3634041/detect-when-android-emulator-is-fully-booted) and all of the various ways of detecting when AVD is booted. `getprop sys.boot_completed` still seems to be the best approach.

But rather than doing the loop/sleep/poll on host machine, I've modified it to do that on the AVD itself.

src/common/AndroidUtils.ts
```diff
@@ -533,33 +536,19 @@ export class AndroidUtils {
...
-
-        return Promise.reject(
-            new Error(`Timeout waiting for emulator-${portNumber} to boot.`)
+        const waitUntilReadyPromise = CommonUtils.promiseWithTimeout(
+            timeout,
+            CommonUtils.executeCommandAsync(command),
+            `Timeout waiting for emulator-${portNumber} to boot.`
         );
+
+        return waitUntilReadyPromise.then(() => Promise.resolve());
```
maliroteh-sf, line 551: This just convert to `Promise<void>`

src/common/AndroidUtils.ts
```diff
@@ -720,6 +709,26 @@ export class AndroidUtils {
...
+            // Ensure value for runtime.network.speed is in lowercase.
+            // Otherwise the device may not be launchable via AVD Manager.
+            const networkSpeed = config.get('runtime.network.speed');
+            if (networkSpeed) {
+                config.set(
+                    'runtime.network.speed',
+                    networkSpeed.trim().toLocaleLowerCase()
+                );
+            }
+
```
maliroteh-sf, line 731: If we create a new AVD using `AndroidUtils.createNewVirtualDevice()` then users cannot launch the AVD from within AVD Manager. They can only be launched from terminal. After a lot of debugging I figured out the issue:

1. A while ago we switched from using `tools/avdmanager` to `cmdline-tools/avdmanager` since it seems that `tools` are apparently "deprecated" in favor of `cmdline-tools` and are not being maintained.
2. When you use `cmdline-tools/avdmanager` to create a new AVD, it created a `config.ini` file for the AVD that has the following 2 entries in it:
```
runtime.network.latency=None
runtime.network.speed=Full
```
3. The problem with the above is that the values should be in lower case (`none`, `full`) otherwise when the user tries to launch the AVD using AVD Manager, it won't launch and instead errors out (but you can still launch the AVD using the `emulator` command from terminal)
4. So here we are explicitly changing those values to lower case.

Leave it to Google to create tools that break their own other tools 🤦

src/common/Requirements.ts
```diff
@@ -113,12 +113,7 @@ export abstract class BaseSetup implements RequirementList {
 
     constructor(logger: Logger) {
         this.logger = logger;
-        this.requirements = [
```
maliroteh-sf, line 116: checking for `lwc-dev-server` should not be part of BaseSetup

src/common/AndroidUtils.ts
```diff
@@ -458,11 +458,14 @@ export class AndroidUtils {
                         );
                     });
                     child.stderr.on('data', (data) => {
-                        reject(
-                            new Error(
-                                `Could not create emulator. Command failed: ${createAvdCommand}\n${data}`
-                            )
-                        );
+                        const msg = data.toString() as string;
```
khawkins, line 461: Just a minor point of a philosophical question here: would it make more sense to type `data` above (to `Buffer`), instead? It feels like that's where the type inference falls down, more than casting to `string` here.

src/common/CommonUtils.ts
```diff
@@ -24,6 +24,23 @@ export class CommonUtils {
         return Promise.resolve();
     }
 
+    public static promiseWithTimeout<T>(
```
khawkins, line 27: That's pretty clever. 🙂

khawkins: But please add a comment block. 😂

khawkins (approved):

src/common/CommonUtils.ts
```diff
@@ -29,6 +29,11 @@ export class CommonUtils {
         promise: Promise<T>,
         failureMessage?: string
     ): Promise<T> {
+        // Javascript/TypeScript do not have promises that timeout. However we could use
```
khawkins, line 32: Oh sorry I just meant the JSDoc. The method made sense to me, but given our eventual mission to JSDoc everything, I was suggesting we make sure we cover it here too.

PR #20, opened 2021-03-08
Add ability to retry a failing ADB command

maliroteh-sf:
1. Add support for retry mechanism to re-execute a failing ADB command
2. Add methods to `AndroidUtils` to easily allow for remounting as root with writable system access
3. Add `apiLevel` to `AndroidVirtualDevice` and the code to fetch its value from an AVDs config file
4. Some cleanup

### DIFF
--- a/src/common/AndroidLauncher.ts
+++ b/src/common/AndroidLauncher.ts
@@ -16,6 +16,19 @@ export class AndroidLauncher {
         this.emulatorName = emulatorName;
     }
 
+    /**
+     * Attempts to preview a Lightning Web Component. It will create the target emulator if it doesn't already exist.
+     * It will launch the emulator and wait for it to boot and will then proceed to preview the LWC. If the preview
+     * target is the browser then it will launch the emulator browser for previewing the LWC. If the preview target
+     * is a native app then it will install & launch the native app for previewing the LWC.
+     *
+     * @param compName Name of the LWC component.
+     * @param projectDir Path to the LWC project root directory.
+     * @param appBundlePath Optional path to the app bundle of the native app. This will be used to install the app if not already installed.
+     * @param targetApp The bundle ID of the app to be launched.
+     * @param appConfig An AndroidAppPreviewConfig object containing app configuration info.
+     * @param serverPort The port for local dev server.
+     */
     public async launchPreview(
         compName: string,
         projectDir: string,
@@ -62,13 +75,7 @@ export class AndroidLauncher {
             })
             .then((actualPort) => {
                 emulatorPort = actualPort;
-                CommonUtils.startCliAction(
-                    `Launching`,
-                    `Waiting for device ${emuName} to boot`
-                );
-                return AndroidUtils.waitUntilDeviceIsReady(emulatorPort);
-            })
-            .then(() => {
+
                 const useServer = PreviewUtils.useLwcServerForPreviewing(
                     targetApp,
                     appConfig
--- a/src/common/AndroidUtils.ts
+++ b/src/common/AndroidUtils.ts
@@ -221,7 +221,7 @@ export class AndroidUtils {
     }
 
     /**
-     * Attempts to fetch a specific Android virtual devices.
+     * Attempts to fetch a specific Android virtual device.
      *
      * @returns An AndroidVirtualDevice object for the specified virtual device, or undefined if device not found.
      */
@@ -665,12 +665,12 @@ export class AndroidUtils {
      *
      * @param portNumber The ADB port of the Android virtual device.
      * @param numRetries Optional parameter that indicates the number of retries if an error occurs. Defaults to 0 (no retries).
-     * @param retryDelay Optional parameter that indicates the milliseconds of delay before retrying again. Defaults to 500 milliseconds.
+     * @param retryDelay Optional parameter that indicates the milliseconds of delay before retrying again. Defaults to AndroidConfig.AdbRetryAttemptDelay.
      */
     public static async remountAsRootWritableSystem(
         portNumber: number,
         numRetries: number = 0,
-        retryDelay: number = 500
+        retryDelay: number = PlatformConfig.androidConfig().AdbRetryAttemptDelay
     ): Promise<void> {
         const emulatorName = await AndroidUtils.fetchEmulatorNameFromPort(
             portNumber
@@ -795,7 +795,7 @@ export class AndroidUtils {
 
     /**
      * Attempts to wait for an Android virtual device to power off. It determines whether a device is powered off
-     * by continously running 'adb devices' command and see if this particular device is no longer in the list.
+     * by continuously running 'adb devices' command and see if this particular device is no longer in the list.
      *
      * @param portNumber The ADB port of the Android virtual device.
      */
@@ -816,7 +816,7 @@ export class AndroidUtils {
 
         return waitUntilReadyPromise.then(() => {
             // It may often be the case that the device is removed from 'adb devices' list too quickly
-            // specially on Windows platform. So we also wait an exra 5 seconds.
+            // specially on Windows platform. So we also wait an extra 5 seconds.
             return CommonUtils.delay(5000);
         });
     }
@@ -827,32 +827,41 @@ export class AndroidUtils {
      * @param command The command to be executed. This should not include 'adb' command itself.
      * @param portNumber The ADB port of the Android virtual device.
      * @param numRetries Optional parameter that indicates the number of times the command will be executed again if ADB comes back with an error. Defaults to 0 (no retries).
-     * @param retryDelay Optional parameter that indicates the milliseconds of delay before retrying again. Defaults to 500 milliseconds.
+     * @param retryDelay Optional parameter that indicates the milliseconds of delay before retrying again. Defaults to AndroidConfig.AdbRetryAttemptDelay.
      */
     public static async executeAdbCommand(
         command: string,
         portNumber: number,
         numRetries: number = 0,
-        retryDelay: number = 500
+        retryDelay: number = PlatformConfig.androidConfig().AdbRetryAttemptDelay
     ): Promise<string> {
         const adbCmd = `${AndroidUtils.getAdbShellCommand()} -s emulator-${portNumber} ${command}`;
 
-        let result: { stdout: string; stderr: string } = {
-            stdout: '',
-            stderr: ''
-        };
+        let allOutput = '';
 
         for (let i = 0; i <= numRetries; i++) {
-            result = await CommonUtils.executeCommandAsync(adbCmd);
+            let result: { stdout: string; stderr: string } = {
+                stdout: '',
+                stderr: ''
+            };
+            let err: any;
+
+            try {
+                result = await CommonUtils.executeCommandAsync(adbCmd);
+            } catch (error) {
+                err = error;
+            }
+
+            allOutput =
+                (err ? err : '') +
+                (result.stderr ? '\n' + result.stderr : '') +
+                (result.stdout ? '\n' + result.stdout : '');
+
             // ADB is terrible and in addition to actual errors it may also send warnings or even success messages to stderr.
             // For example if you run the command 'adb shell reboot -p' to power down a device, when done it prints the success
-            // message 'Done\n' to stderr. And sometimes ADB prints error messages to stdout instead of stderr.
-            if (
-                (result.stderr &&
-                    result.stderr.toLowerCase().includes('error:')) ||
-                (result.stdout &&
-                    result.stdout.toLowerCase().includes('error:'))
-            ) {
+            // message 'Done\n' to stderr. And sometimes ADB prints error messages to stdout instead of stderr. So we first gether
+            // all outputs into one and then explicitly check to see if an error has occured or not.
+            if (allOutput && allOutput.toLowerCase().includes('error:')) {
                 this.logger.warn(
                     `ADB command '${adbCmd}' failed. Retrying ${
                         numRetries - i
@@ -868,7 +877,7 @@ export class AndroidUtils {
         }
 
         // If we got here then it means that the command failed to execute successfully after all retries (if any)
-        return Promise.reject(new Error(result.stderr));
+        return Promise.reject(new Error(allOutput));
     }
 
     /**
--- a/src/common/AndroidUtils.ts
+++ b/src/common/AndroidUtils.ts
@@ -20,9 +20,6 @@ import { PlatformConfig } from './PlatformConfig';
 import { LaunchArgument } from './PreviewConfigFile';
 import { PreviewUtils } from './PreviewUtils';
 
-const spawn = childProcess.spawn;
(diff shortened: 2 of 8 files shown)


src/common/AndroidLauncher.ts
```diff
@@ -62,13 +75,7 @@ export class AndroidLauncher {
             })
             .then((actualPort) => {
                 emulatorPort = actualPort;
-                CommonUtils.startCliAction(
-                    `Launching`,
-                    `Waiting for device ${emuName} to boot`
-                );
-                return AndroidUtils.waitUntilDeviceIsReady(emulatorPort);
-            })
-            .then(() => {
```
maliroteh-sf, line 71: `AndroidUtils.startEmulator()` now takes a param (defaulting to true) indicating whether it should wait for the simulator and will internally call `waitUntilDeviceIsReady` so no need to do this here anymore.

src/common/CommonUtils.ts
```diff
@@ -234,6 +234,24 @@ export class CommonUtils {
         );
     }
 
+    /**
+     * Launches the desktop browser and navigates to the provided URL.
+     *
+     * @returns A Promise that launches the desktop browser and navigates to the provided URL.
+     */
+    public static async launchUrlInDesktopBrowser(url: string): Promise<void> {
```
maliroteh-sf, line 242: This is used in other repos but I thought to bring it here since it is a useful utility function, and instead call into it from other repos.

src/common/IOSLauncher.ts
```diff
@@ -66,13 +79,6 @@ export class IOSLauncher {
                 );
                 return IOSUtils.bootDevice(deviceUDID);
             })
-            .then(() => {
-                CommonUtils.startCliAction(
-                    `Launching`,
-                    `Waiting for device ${deviceUDID} to boot`
-                );
-                return IOSUtils.waitUntilDeviceIsReady(deviceUDID);
-            })
```
maliroteh-sf, line 75: Similar to `AndroidUtils`, the function `IOSUtils..bootDevice() `now takes a param (defaulting to true) indicating whether it should wait for the simulator and will internally call `waitUntilDeviceIsReady` so no need to do this here anymore.

src/common/AndroidUtils.ts
```diff
@@ -504,29 +529,249 @@ export class AndroidUtils {
...
                     );
-                    return Promise.resolve(portNumber);
+
+                    const isWritable = await AndroidUtils.isEmulatorSystemWritable(
+                        portNumber
+                    );
+
+                    if (isWritable === writable) {
+                        // already launched and modes match so just return its port
+                        return Promise.resolve(portNumber);
```
maliroteh-sf, line 539: The logic here ensure that if an emulator is already running then it must be running in the same "mode" as what is passed to this function. So if it is running with `writable-system` but this function is called with `writable=false` then that's a mismatch and we would shut down & relaunch.

Should we leave it this way or should we only care for the scenario where `writeable=true` but `isWritable=false`? In other words if an emulator is running with `writable-system` but this function is called with `writable=false` should we ignore and continue or should I do what I do now and shut it down + relaunch?

src/common/AndroidUtils.ts
```diff
@@ -536,20 +781,96 @@ export class AndroidUtils {
...
+        const waitUntilReadyPromise = CommonUtils.promiseWithTimeout(
+            timeout,
+            CommonUtils.executeCommandAsync(command),
+            `Timeout waiting for emulator-${portNumber} to power off.`
+        );
+
+        return waitUntilReadyPromise.then(() => {
+            // It may often be the case that the device is removed from 'adb devices' list too quickly
+            // specially on Windows platform. So we also wait an exra 5 seconds.
+            return CommonUtils.delay(5000);
```
maliroteh-sf, line 820: This 5sec is measured totally unscientifically 😃  On my Mac & Windows laptops it seems to work just fine.

src/common/AndroidUtils.ts
```diff
@@ -504,29 +529,249 @@ export class AndroidUtils {
...
+            portNumber
+        ).then((result) => result.split('\n')[0].trim());
+    }
+
+    /**
+     * Remounts adb as root with writable system access for the AVD that is running on the specified port. If the AVD currently
+     * is not launched with writable system access, this function will restart it with write access first then remounts as root.
+     *
+     * @param portNumber The ADB port of the Android virtual device.
+     * @param numRetries Optional parameter that indicates the number of retries if an error occurs. Defaults to 0 (no retries).
```
maliroteh-sf, line 667: In the interest of performance I've decided to default numRetires=0 unless a caller explicitly sends a non-zero value. I'm open to changing default value to something else (like 3 tries). I thought to leave it like this first and see if we still see adb command failures in the wild before increasing this number.

src/common/AndroidUtils.ts
```diff
@@ -498,27 +523,47 @@ export class AndroidUtils {
...
+                        );
+
+                        const isWritable = await AndroidUtils.isEmulatorSystemWritable(
+                            portNumber
+                        );
+
+                        if (isWritable === writable) {
+                            // already launched and modes match so just return its port
+                            return Promise.resolve(portNumber);
+                        }
```
maliroteh-sf, line 543: The logic here ensure that if an emulator is already running then it must be running in the same "mode" as what is passed to this function. So if it is running with `writable-system` but this function is called with `writable=false` then that's a mismatch and we would shut down & relaunch.

Should we leave it this way or should we only care for the scenario where `writeable=true` but `isWritable=false`? In other words if an emulator is running with `writable-system` but this function is called with `writable=false` should we ignore and continue or should I do what I do now and shut it down + relaunch?

src/common/AndroidUtils.ts
```diff
@@ -223,6 +220,32 @@ export class AndroidUtils {
             });
     }
 
+    /**
+     * Attempts to fetch a specific Android virtual devices.
```
sfdctaka, line 224: devices -> device

src/common/AndroidUtils.ts
sfdctaka: Can you remind me why we are doing this? I remember last week you found that any modification done to `writable-system` gets wiped when the emulator is restarted without that flag. So wouldn't it be to our benefit to just let it keep running as a `writable-system`? Or you're trying to respect what the command says, meaning, if the configuration is different from previous run then shut down & relaunch?

maliroteh-sf: Yes that's exactly the case. I'm currently trying to respect the parameters of the command simply b/c what we have in core should be generic to be used by all repos and not tailored for a specific use case.

src/common/AndroidUtils.ts
```diff
@@ -536,20 +781,96 @@ export class AndroidUtils {
...
+            AndroidUtils.executeAdbCommand(command, portNumber),
             `Timeout waiting for emulator-${portNumber} to boot.`
         );
 
         return waitUntilReadyPromise.then(() => Promise.resolve());
     }
 
+    /**
+     * Attempts to wait for an Android virtual device to power off. It determines whether a device is powered off
+     * by continously running 'adb devices' command and see if this particular device is no longer in the list.
```
sfdctaka, line 798: continously -> continuously

src/common/AndroidUtils.ts
```diff
@@ -536,20 +781,96 @@ export class AndroidUtils {
...
+
+        const waitUntilReadyPromise = CommonUtils.promiseWithTimeout(
+            timeout,
+            CommonUtils.executeCommandAsync(command),
+            `Timeout waiting for emulator-${portNumber} to power off.`
+        );
+
+        return waitUntilReadyPromise.then(() => {
+            // It may often be the case that the device is removed from 'adb devices' list too quickly
+            // specially on Windows platform. So we also wait an exra 5 seconds.
```
sfdctaka, line 819: exra -> extra

src/common/AndroidUtils.ts
sfdctaka: Number of retries in `remountAsRootWritableSystem` and the delay here. I wonder if we should make it configurable through JSON or environment variable. 5 seconds is reasonable to me. Just thinking about the case if user wants to take control.

sfdctaka (approved):
Few typos. Overall LGTM!

src/common/AndroidUtils.ts
maliroteh-sf: For now we are the only ones that are consumers of this code and we can control those values through the parameters to `remountAsRootWritableSystem`. As for this 5sec delay I don't think we should expose it b/c it is really a bad thing that we're doing here. I still would like to investigate other options and eventually remove this 5sec sleep from our code.

src/common/AndroidUtils.ts
khawkins: > Should we leave it this way or should we only care for the scenario where writeable=true but isWritable=false?

I think we should only care about the case where we need a writeable file system but it's not booted that way. We can come back and change it later if we need to, but our requirement right now is simply requiring a writeable file system when we require it.

src/common/AndroidUtils.ts
khawkins: I feel like these values should be configured in code (or outside of it, if we feel that's necessary), and not otherwise passed in as parameters. It's logic internal to the method itself—it shouldn't drive the API of the method.

maliroteh-sf: @khawkins There are only 2 functions that use `numRetries` and `retryDelay`. These 2 are `executeAdbCommand` and `remountAsRootWritableSystem`.

For `remountAsRootWritableSystem` we could take the parameters out from function signature and put these inside its code. But I didn't do that b/c for `executeAdbCommand` I think it should take them as parameters. There are many functions that call into `executeAdbCommand` such as:

```
stopEmulator() // calls it to shut down an AVD
rebootEmulator() // calls it to reboot an AVD
isEmulatorSystemWritable() // calls it to get the path to the AVD folder on host machine
fetchEmulatorNameFromPort() // calls it get the name from AVD running on a port
waitUntilDeviceIsReady() // calls it to run shell on an AVD and wait for device
launchURLIntent() // calls it to open url
launchNativeApp() // calls it to install & launch an app
remountAsRootWritableSystem() // calls it to run a series of commands
```

With the exception of the last one (which is where we actually observed adb issues) none of the other ones need their commands to be retried.

So either we should have the parameters for `executeAdbCommand` OR have 2 versions of it `executeAdbCommand` and `executeAdbCommandWithRetry` and then have `executeAdbCommandWithRetry` define these inside its code instead of taking them as parameters (but then that will lead to code duplication).

khawkins: It feels like `executeAdbCommandWithRetry` *should* be its own method, calling `executeAdbCommand` internally, but managing all of the "if error message retry" logic itself. It's a subtle distinction, but it doesn't feel like `executeAdbCommand` should be carrying that burden when it's not really its responsibility, at the end of the day.

I guess where that doesn't work is the fact that `executeAdbCommand` currently encapsulates the I/O aspects of its responsibilities and only returns output.  I guess it's fine to leave it as is.

src/common/AndroidUtils.ts
```diff
@@ -527,6 +572,206 @@ export class AndroidUtils {
...
+
+    /**
+     * Remounts adb as root with writable system access for the AVD that is running on the specified port. If the AVD currently
+     * is not launched with writable system access, this function will restart it with write access first then remounts as root.
+     *
+     * @param portNumber The ADB port of the Android virtual device.
+     * @param numRetries Optional parameter that indicates the number of retries if an error occurs. Defaults to 0 (no retries).
+     * @param retryDelay Optional parameter that indicates the milliseconds of delay before retrying again. Defaults to AndroidConfig.AdbRetryAttemptDelay.
+     */
+    public static async remountAsRootWritableSystem(
```
khawkins, line 670: I'm a little confused by the purpose of this method. It feels like we have half of the "boot a writeable emulator" code in `AndroidUtils.startEmulator()`, and the other half of it in here. Without seeing the consumption, I'm having a hard time ascertaining why it doesn't all just go in some form in `AndroidUtils.startEmulator()`.

src/common/AndroidUtils.ts
khawkins: Me too. 😀 It's an area that would be great if we could react to whatever adversity exists on the other side (e.g. bringing it back up), and employing retry logic there too.

I'm thinking there's going to come a day where we want to retool this flow. Some of the centralization of code (e.g. the bullet-proofing in `executeAdbCommand()`) may actually be limiting us, and it might make more sense to decentralize / bubble up the `adb` failure fallout into individual methods, which can then tailor their responses to their specific scenarios.

Sure would be great if we had a reliable API for these actions. 😐

src/common/IOSUtils.ts
```diff
@@ -18,15 +18,33 @@ const RUNTIME_TYPE_PREFIX = 'com.apple.CoreSimulator.SimRuntime';
 const LOGGER_NAME = 'force:lightning:local:iosutils';
 
 export class IOSUtils {
+    /**
+     * Initialized the logger used by AndroidUtils
```
khawkins, line 22: `AndroidUtils` is a copy/paste error here.

khawkins (approved):
Looks fine over all. Dealing with `adb` is definitely an imperfect science.

src/common/AndroidUtils.ts
khawkins: I guess I'm not confused by the method itself, but rather more specifically why we're putting some of the "make sure it's a writeable partition" logic in here.

maliroteh-sf: I understand exactly what you're referring to and was struggling myself too, to choose an approach over the other. At the end of the day the reasons I choose this approach were:

1. I initially had `remountAsRootWritableSystem()` take parameters for defining retry/delay values and didn't want those to be added to `startEmulator()` (this is not the case anymore since as per your feedback I have removed the parameters and move them to `executeAdbCommandWithRetry()`)

2. Putting the logic of this method into `startEmulator()` would make that method too large/complex.

3. I wasn't sure if there ever is a case for launching an emulator with `-writable-system` but not needing to also run `adb root && adb remount`. In my view the methods in `AndroidUtils` should be generic enough to serve many needs and not just tailored for specific use cases. To me, merging `remountAsRootWritableSystem` and `startEmulator()` together seemed like reducing the generality of `AndroidUtils` to better serve a specific use case. Whereas leaving them separate would keep things generic enough and would allow the consumer to decide whether they want to also `adb root && adb remount` on top of `-writable-system` (specially since for API >= 29 this would also involve disabling ABV and Verify). Again this is b/c I wasn't sure if there ever is a case for launching an emulator with `-writable-system` but not needing to also run `adb root && adb remount`. If we feel that (in addition to our current use cases) most if not all of the times, `-writable-system` will be used in conjunction with `adb root && adb remount` then I can go ahead and merge the two.

khawkins: I think we can leave it for now. I too don't want any given method to get super unwieldy or carry too many responsibilities. Let's keep it as it is.

PR #22, opened 2021-03-11
show CLI messages when starting/remounting emulators

maliroteh-sf:


src/common/AndroidUtils.ts
```diff
@@ -436,7 +436,7 @@ export class AndroidUtils {
         abi: string
     ): Promise<void> {
         // Just like Android Studio AVD Manager GUI interface, replace blank spaces with _ so that the ID of this AVD
-        // doesn't have blanks (since that's not allowed). AVD Manager will automatially replace _ back with blank
```
maliroteh-sf, line 439: Spell Checker extension is finding all of my mistakes now 😝

src/common/AndroidUtils.ts
```diff
@@ -510,45 +510,57 @@ export class AndroidUtils {
...
                     if (writable === false || isWritable === true) {
                         // If we're not asked for a writable, or if it already
                         // is writable then we're done so just return its port.
-                        return Promise.resolve(resolvedPortNumer);
+                        return Promise.resolve(resolvedPortNumber);
                     }
 
                     // mismatch... shut it down and relaunch it in the right mode
-                    await AndroidUtils.stopEmulator(resolvedPortNumer);
+                    CommonUtils.updateCliAction(
```
maliroteh-sf, line 530: leaving these as hard-coded strings for now b/c we have it that way in all 3 repos currently and have a separate story about cleaning that up.

src/common/AndroidUtils.ts
```diff
@@ -510,45 +510,57 @@ export class AndroidUtils {
...
+                        'emulator currently launched without the -writable-system flag so shutting it down first'
```
khawkins, line 531: Minor: I propose a little more concise for a status message:

> Emulator currently launched without -writable-system flag. Shutting down.

src/common/AndroidUtils.ts
```diff
@@ -510,45 +510,57 @@ export class AndroidUtils {
...
+                    );
+                    await AndroidUtils.stopEmulator(resolvedPortNumber);
                 }
 
+                CommonUtils.updateCliAction(
+                    `launching ${resolvedEmulatorName} on port ${resolvedPortNumber}${
```
khawkins, line 537: Can we conditionally add "Re-" to this, based on whether it had previously been running?

src/common/AndroidUtils.ts
khawkins: Also, can we start all of these messages with capital letters? We're a little inconsistent with our format, across repos, but if we're going to standardize (which I think we should), I'd prefer they start like a sentence.

src/common/AndroidUtils.ts
maliroteh-sf: sure, will do

src/common/AndroidUtils.ts
```diff
@@ -725,10 +750,14 @@ export class AndroidUtils {
 
                 return Promise.resolve();
             })
-            .then(() =>
-                // Now we're ready to remount and truely have root & writable access to system
-                AndroidUtils.executeAdbCommandWithRetry('remount', portNumber)
-            )
+            .then(() => {
+                CommonUtils.updateCliAction('remounting');
```
khawkins, line 754: Can we add more detail? "Remounting system partition to be writable"

khawkins (approved):
Looks good over all. I think these messages will help.

src/common/AndroidUtils.ts
maliroteh-sf: Sure I can do that. The reason I didn't is b/c the format would be `Action Message...Status Message` for example:

```
Starting device... emulator currently launched without -writable-system. Shutting down.
Starting device... relaunching sfdxdebug on port 5572 with -writable-system
Starting device... waiting for sfdxdebug to boot
Starting device... done
```

The last one (`done`) is automatically printed in lowercase when you call `cli.action.stop` so because of that I kept all of these as lowercase too.

khawkins: Ah yes I forgot about the overall message format too. Yeah, that's fine then.

PR #25, opened 2021-03-15
add more jsdoc comments

maliroteh-sf:


src/common/AndroidTypes.ts
```diff
@@ -80,6 +86,10 @@ export class AndroidPackages {
     public platforms: AndroidPackage[] = [];
     public systemImages: AndroidPackage[] = [];
 
+    /**
+     * Checks to see if the object is empty (i.e the platforms and system images are both empty)
+     * @returns True if empty, False otherwise.
```
sfdctaka, line 91:
```suggestion
     * @returns True if empty, false otherwise.
```

nit

sfdctaka (approved):
LGTM!

khawkins (approved):

src/common/Requirements.ts
```diff
@@ -39,11 +39,15 @@ export interface RequirementList {
...
-// This function wraps existing promises with the intention to allow the collection of promises
-// to settle when used in conjunction with Promise.all(). Promise.all() by default executes until the first
-// rejection. We are looking for the equivalent of Promise.allSettled() which is scheduled for ES2020.
-// Once the functionality is  available  in the near future this function can be removed.
-// See https://github.com/tc39/proposal-promise-allSettled
+/**
+ * This function wraps existing promises with the intention to allow the collection of promises
+ * to settle when used in conjunction with Promise.all(). Promise.all() by default executes until the first
+ * rejection. We are looking for the equivalent of Promise.allSettled() which is scheduled for ES2020.
+ * Once the functionality is  available  in the near future this function can be removed.
```
rwhitley, line 46: When the functionality is available this function can be removed.

(Do you really want to promise "near future"?) :-)

src/common/Requirements.ts
```diff
@@ -39,11 +39,15 @@ export interface RequirementList {
...
+ * See https://github.com/tc39/proposal-promise-allSettled
+ * @param requirement A requirement object
```
rwhitley, line 48: A Requirement object

src/common/Requirements.ts
```diff
@@ -116,6 +120,10 @@ export abstract class BaseSetup implements RequirementList {
         this.additionalRequirements = [];
     }
 
+    /**
+     * Executes all of the base and additional requirement steps.
+     * @returns A SetupTestResult containing all of the result of the requirement tests
```
rwhitley, line 125: "results"?

src/common/Requirements.ts
```diff
@@ -201,12 +209,20 @@ export abstract class BaseSetup implements RequirementList {
...
+     */
     public addBaseRequirements(reqs: Requirement[]) {
         if (reqs) {
             this.requirements = this.requirements.concat(reqs);
         }
     }
 
+    /**
+     * Adds a new additional requirement to the list of the additional requirements to be verified.
+     * @param reqs The new additional requirement to be added.
```
rwhitley, line 224: Appears to be an array of Requirement objects...

rwhitley (commented):
I was nearly finished before I realized this was submitted on core! Sorry to intrude (and to have done so after the fact).

src/common/AndroidTypes.ts
```diff
@@ -258,6 +274,9 @@ export class AndroidVirtualDevice {
         this.apiLevel = apiLevel;
     }
 
+    /**
+     * A string representation of an AndroidVirtualDevice which includes Display Name, Device Name, API
```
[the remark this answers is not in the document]
maliroteh-sf: No it's actually API and not API level (although they're related). API would be strings like "Android 9.0 Pie" where as API level for that string would be "API 28". But I know that in different parts of AVD Manager instead of API they call it Target so maybe I'll rename this to be Target.

src/common/Common.ts
```diff
@@ -110,14 +143,29 @@ export class Version {
         this.patch = patch;
     }
 
+    /**
+     * Verifies that the input version is an exact match.
```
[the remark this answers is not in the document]
maliroteh-sf: Oh this has nothing to do with Android. It's just a class that defines semantic versioning. And the way this class and its methods are defined, it would be used in cases like:

```
let v1 = new Version(....)
let v2 = new Version(....)
if (v1.same(v2)) {
. . . 
}
```

So methods like `same`,  `sameOrNewer`, etc will be invoked on an instance and take another instance. This is something common that developers are very well familiar with.

PR #31, opened 2021-03-25
@W-8888184 - Clean up and consolidate requirement data structures.

sfdctaka:
Decoupling `Requirement`-related classes and `Setup` command so that other commands don't need to inherit from `Setup`.

What you'll find here:
* Renaming of data structures where I felt appropriate, like `BaseSetup` to `CommandRequirement`, so to break the mental linkage one might have(I did)
* Composable by requirements per `SfdxCommand` through `CommandRequirement`'s properties as described [here](https://salesforce.quip.com/oPpPA07RHQR4).
* `executeSetup()` will no longer return `SetupTestResults` as it serves no purpose.
* Removed "setup" trails that unnecessarily relates `Setup` to requirement related classes.


sfdctaka:
@maliroteh-sf @khawkins This is the approach that I like to take in `core` foundational change to decouple requirement and Setup command. I would be grateful to get your early feedback before I keep moving on. Thanks!

src/common/__tests__/AndroidEnvironmentSetup.test.ts
```diff
@@ -40,7 +40,7 @@ describe('Android enviroment setup tests', () => {
     let androidEnvironment: AndroidEnvironmentSetup;
 
     beforeEach(() => {
-        androidEnvironment = new AndroidEnvironmentSetup(logger);
+        androidEnvironment = new AndroidEnvironmentSetup(logger, 'Android');
```
maliroteh-sf, line 43: Use `CommandLineUtils.ANDROID_FLAG` instead of hardcoding here

src/common/__tests__/IOSEnvironmentSetup.test.ts
```diff
@@ -57,7 +57,7 @@ describe('IOS Environment Setup tests', () => {
     let iosEnvironment: IOSEnvironmentSetup;
 
     beforeEach(() => {
-        iosEnvironment = new IOSEnvironmentSetup(logger);
+        iosEnvironment = new IOSEnvironmentSetup(logger, 'iOS');
```
maliroteh-sf, line 60: Use `CommandLineUtils.IOS_FLAG` instead of hardcoding here

src/common/Requirements.ts
```diff
@@ -116,47 +118,52 @@ export abstract class BaseSetup implements RequirementList {
...
+        this.commandRequirements = { requirements: [], enabled: true };
+        this.requirementsCheckFailureMessage = '';
+        this.requirementsCheckRecommendationMessage = '';
     }
 
     /**
      * Executes all of the base and additional requirement steps.
-     * @returns A SetupTestResult containing all of the results of the requirement tests
      */
-    public async executeSetup(): Promise<SetupTestResult> {
```
maliroteh-sf, line 127: Since we're not returning `SetupTestResult` anymore, then does the interface definition for it still needs to be using `export`? Same for `RequirementResult` interface

maliroteh-sf (commented):
Overall this looks good to me so far and I've left some comments.

I know that we're doing away with the execute method returning the test results and I'm ok with that. The only place we used to use that was inside the `Setup` command in order to resolve/reject based on the test results but now that the `Requirement` class is doing that handling then I don't see any reason for it returning the test results. In the future if we see the use-case for it then we can add it back and return it.

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "resolveJsonModule": true,
     "pretty": true,
     /* Basic Options */
-    "target": "esnext",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
+    "target": "es2019",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
```
sfdctaka, line 7: I like to use Typescript's optional chaining so targeting `es2019` instead. With `esnext` transpiler doesn't handle optional chaining correctly.

src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -50,40 +48,13 @@ export class Setup extends SfdxCommand {
...
-                                    'error:setupFailed:recommendation'
-                                )
-                            ]
-                        )
-                    );
-                } else {
-                    return Promise.resolve(result);
-                }
-            });
+            .then(() => this.setup().executeSetup()); // verify requirements
```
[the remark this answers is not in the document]
sfdctaka: I went with `executeChecks` because it is the checks for requirements that runs.

src/common/Requirements.ts
sfdctaka: I went with `RequirementCheckResult`.

src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -50,40 +48,13 @@ export class Setup extends SfdxCommand {
...
-    public skipBaseRequirements = false;
-    public skipAdditionalRequirements = false;
-
-    private setupSteps: BaseSetup | undefined;
-
-    public async run(direct: boolean = false): Promise<any> {
-        if (direct) {
-            await this.init(); // ensure init first
-        }
+    public async run(): Promise<any> {
```
maliroteh-sf, line 51: Since you're changing the method signature for the `run()` method, you'll need to do that for the other commands in other repos too.

I'm assuming you were going to update all other repos too since we don't want our commands to extend from `Setup` anymore (which they currently are).

src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -119,7 +86,7 @@ export class Setup extends SfdxCommand {
         if (this.flags.apilevel) {
             if (CommandLineUtils.platformFlagIsIOS(this.flags.platform)) {
                 this.logger.warn(
-                    'The apiLevel flag does not apply to the iOS platform... ignoring.'
+                    messages.getMessage('error:doesNotApplyToIOS')
```
maliroteh-sf, line 89: nit: This is a warning not an error so perhaps it should be `warning:doesNotApplyToIOS`. Also in my previous effort to take hardcoded strings to message files I didn't bother take the hardcoded strings that we log b/c they are not exposed to the user. They will be in logs / log files instead.

src/common/Requirements.ts
```diff
@@ -22,21 +23,21 @@ export interface Requirement {
...
     title: string;
 }
 
-export interface SetupTestResult {
+interface RequirementCheckResult {
     hasMetAllRequirements: boolean;
     tests: RequirementResult[];
 }
 
 export interface RequirementList {
```
maliroteh-sf, line 38: does this need an `export`?

maliroteh-sf (commented):
FYI, the team practice is that we're supposed to create a fork of this repo (and the other repos), make changes to our fork and then create a PR from our fork into this repo. I `think` if you don't do that then the PR jobs won't kick in to verify a build. We can manually verify this one ourselves but for future PRs please keep that in mind.

sfdctaka:
Yikes 😬 . I was under the impression that forking was only necessary for `lwc-dev-mobile` only. Once I get approvals from you and Kevin I can do that so the merge can take place through official passage.

src/common/Requirements.ts
```diff
@@ -103,60 +104,88 @@ export function WrappedPromise(
...
-    public skipBaseRequirements: boolean = false;
-    public skipAdditionalRequirements: boolean = false;
-    public requirements: Requirement[];
-    public additionalRequirements: Requirement[];
+export class CommandRequirement {
+    public baseRequirements: RequirementList = {
+        requirements: [],
+        enabled: true
+    };
+    public commandRequirements: RequirementList = {
```
khawkins, line 112: The separation of `baseRequirements` and `commandRequirements` is unclear to me. How will this differentiation work in practice? I had assumed we were going with a model of distinct requirements lists, and this would seem to indicate otherwise.

src/common/Requirements.ts
```diff
@@ -103,60 +104,88 @@ export function WrappedPromise(
         });
 }
 
-export abstract class BaseSetup implements RequirementList {
-    public skipBaseRequirements: boolean = false;
-    public skipAdditionalRequirements: boolean = false;
-    public requirements: Requirement[];
-    public additionalRequirements: Requirement[];
+export class CommandRequirement {
```
khawkins, line 107: Minor: should we consider `CommandRequirements` (with an 's') instead? By its makeup, it's always `1..n` requirements.

src/common/Requirements.ts
sfdctaka: I may have misinterpreted what was on the spec. This is pseudo-code from the spec:
image
`commandRequirements` is what I wanted to emulate like `createRequirements` in the screenshot. My thinking was you probably wanted others like `hostRequirements` or `proxyRequirements`. Instead of specific names I am using the generic `commandRequirements`. 
So a command is composed with 2 requirements, `baseRequirements` which it can be used to compose prerequisite checks and `commandRequirements` which can focus on the command's specific requirement checks.

src/common/Requirements.ts
sfdctaka: The reason I didn't make it plural was because the class already has a property called `commandRequirements`. I also wanted the developer using this class be able to something like
`let commandRequirements: CommandRequirement[] = [];`
where 's' in the name of the variable can indicate a collection.
Maybe the name `CommandRequirement` here is not a good name to start with. I can add 's' or change the name. Maybe `CommandChecks`?

sfdctaka:
Closing this PR and continuing in a new one from my fork.

PR #33, opened 2021-03-31
@W-8888184 - Clean up and consolidate requirement data structures.

sfdctaka:
Decoupling Requirement-related classes and Setup command so that other commands don't need to inherit from Setup.

What you'll find here:
* Renaming of data structures where I felt appropriate. No more `BaseSetup` that was confusing(at least for me).
* `Requirements` are composable per `SfdxCommand` via`CommandRequirements` property, that needs to be added per `interface HasRequirements`.
* `executeSetup()` is now simply `execute()`. It will no longer return `SetupTestResults` as it serves no purpose.
* Removed "setup" trails that unnecessarily relates `Setup` command to any `Requirement`-related classes.


src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -145,21 +112,15 @@ export class Setup extends SfdxCommand {
...
                 this.flags.platform
             )
-                ? new AndroidEnvironmentSetup(this.logger, this.flags.apilevel)
-                : new IOSEnvironmentSetup(this.logger);
+                ? new AndroidEnvironmentChecks(this.logger, this.flags.apilevel)
+                : new IOSEnvironmentChecks(this.logger);
         }
 
-        return this.setupSteps;
+        return this._commandChecks;
```
maliroteh-sf, line 124: I'm having a hard time trying to understand how the refactoring in this PR is going to play out in other repos. Since we don't want our commands to extent from `Setup` command anymore, does it mean that every command is now going to duplicate something similar to this code snippet? And how can we determine if a command should run the environment checks first or should just skip them?

sfdctaka: We can't enforce coding pattern through inheritance. This was what made `Setup` a culprit in hoarding requirements in the firsts place. I can look into ways so other commands can be coded within the guard rails and have well-known pattern to follow to have their own set of requirements. However, if a rigid standard is going to be brittle or come up short I would rather offer flexibility and have the coder of the sfdx command to opt-in to having requirements run, either through the approach shown here, or, any approach that he or she can come up with. If you have suggestion for putting a system in place for it I like to take a stab at that too.

maliroteh-sf: Well at this point I just would like to understand how the current approach is going to work in different repos. Our commands may be executed individually from the command line, or in tandem (e.g from the uber command). In the case of the former, we need to have a way for a command to execute the setup checks first and then do its thing, whereas for the latter we want a command to skip the check. Right now I don't quite clearly see how we will accomplish this after the refactor. But then again once we start consuming it in the other repos, it might become more clear.

khawkins: I'm going to add some commentary that spans different code bodies in this PR, but has roots in this conversation too, so I'm dumping it all here. 🙂

1. But we'll start here. I'm not a fan of every Command having its own `commandChecks` property. The requirements processor (`CommandChecks`) can effectively be represented as a static processor. It takes requirements, it executes them, it reports results. This should be a service, not tied to a particular object that needs to use it.
2. As a corollary but also a point of allocation of responsibility, `Checks` should be owned by Commands, not a `CommandChecks` process. The latter should take them as an input.
3. I'd personally prefer to keep the naming of the base objects around the concept of requirements. `CommandChecks` could be `RequirementsProcessor` or similar. `Checks` could be `RequirementsCollection` or something like that. It's all related to requirements, and I feel like the tack to "checks" unnecessarily divorces that context.

src/common/Requirements.ts
```diff
@@ -22,21 +23,25 @@ export interface Requirement {
...
 }
 
-export interface RequirementList {
+interface RequirementList {
     requirements: Requirement[];
-    executeSetup(): Promise<SetupTestResult>;
+    enabled: boolean;
+}
+
+interface Checks {
```
khawkins, line 43: In addition to my previous comments about requirements groups being owned by commands, I'd propose that this interface change a bit, to formalize that relationship by having those commands implement this interface.  So I'm thinking this becomes more like:
```typescript
interface HasRequirements {
    commandRequirements: { [key: string]: RequirementList };
}
```
And then your class definition becomes e.g.:
```typescript
export class Setup extends SfdxCommand implements HasRequirements { }
```
That way each command has the same property name with the same definition, for exposing these requirements.

src/cli/commands/force/lightning/local/setup.ts
maliroteh-sf: I like the idea of having a service... but I think taking that approach would turn out to be much more involved and require more work. Right now `Requirements.execute()` runs through an array of requirements **and** prints out the results. If we move to a service approach then I think we should do the right thing and have the service only be in charge of running the checks + returning their results, and have the caller do the printing. I don't think the service should be in charge of printing the results on the screen.... but then again the way we want things to work is that the checks should be running in parallel and we print the results of a check as soon as it is finished. Maybe we would need use Observables from RxJS or just have a dirty service that runs the tasks and prints the results even though it shouldn't be printing the results.... The refactoring task is quickly becoming something really big.

khawkins: I agree in principle with the concept of a properly-partitioned service, like you describe. As an intermediate step, I was hoping to simply take the code we have and effectively make it standalone.

This is basically what the `BaseSetup` class does today. I'm proposing that our first refactor simply separate that engine out from the requirements themselves.

sfdctaka:
@maliroteh-sf @khawkins I created `RequirementProcessor`. It looks much, much cleaner now with that level of indirection :)

src/common/Requirements.ts
```diff
@@ -22,21 +23,27 @@ export interface Requirement {
...
     title: string;
 }
 
-export interface SetupTestResult {
+interface RequirementCheckResult {
     hasMetAllRequirements: boolean;
     tests: RequirementResult[];
 }
 
 export interface RequirementList {
```
maliroteh-sf, line 38: Looks like we're not using `RequirementList` outside this file so perhaps we should remove `export`

src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -50,40 +53,24 @@ export class Setup extends SfdxCommand {
...
+                RequirementProcessor.execute(
+                    this.logger,
+                    this.commandRequirements,
+                    requirementMessages.getMessage(
+                        'error:requirementCheckFailed'
+                    ),
+                    requirementMessages.getMessage(
+                        'error:requirementCheckFailed:recommendation'
+                    )
+                )
```
maliroteh-sf, line 72: This doesn't make much sense to me. We're importing `requirementMessages` from `Requirements.ts` and then passing it back to `RequirementProcessor.execute()` in the same file. I understand that we want the `execute()` method to take messages for failure+remediation so I think we should:

1. define these 2 messages in `setup.json` message file instead and don't import `requirementMessages`

Or

2. If we suspect that the value of the 3rd and 4th parameters of `RequirementProcessor.execute()` would always be `error:requirementCheckFailed` and `error:requirementCheckFailed:recommendation` regardless of which command is calling `RequirementProcessor.execute()` then why not just remove the 3rd and 4th parameters  altogether and just have that method directly use those messages.

maliroteh-sf (commented):
Overall looks good.

src/common/Requirements.ts
sfdctaka: `RequirementList` is used by `AndroidEnvironmentRequirements` and `IOSEnvironmentRequirements`.

src/cli/commands/force/lightning/local/setup.ts
sfdctaka: `requirementMessages` is not imported but declared and initialized in `Requirements.ts`. It is also exported for use by other files(`AndroidEnvironmentRequirements` and `IOSEnvironmentRequirements`). It is referenced in `execute()` method but not passed in.

`requirement.json` is already sprinkled with resource strings used by both `AndroidEnvironmentRequirements` and `IOSEnvironmentRequirements`. Maybe this is the source of the evil 😈 

For option 1 are you suggesting that the other strings besides `error:requirementCheckFailed` and `error:requirementCheckFailed:recommendation` should be moved as well? On this note, yes, the resource file(`setup.json`) had also been a victim of it getting conflated from coupling of `Setup` and `Requirement` and needs even more careful distribution of resource strings between itself and `requirement.json` than my attempt here.

For option 2, if the caller of `RequirementProcessor.execute()` were to determine whether to show failure&recommendation messages or not we would should decide if the return value of `Promise.reject()` suffices or revive returning `SetupTestResult` so that the caller has a chance to construct failure&recommendation messages from the failed tests should he chooses to. The reason why I opted to pass failure&recommendation messages as parameters to `execute()` was because earlier we decided that we won't return `SetupTestResult` no more from `exeucte()`. And, the caller in that case should be clearly aware if any of the requirement checks fail they know what messages they want to convey to the user.

I'm open to either way. What I do want to do though and get yours and Kevin's opinion is removing the `logger` from `execute()`. The only time it is used is in `catch` handler.

maliroteh-sf: I may be wrong but I **think** when we consume -core in other repos, other commands would also make this exact call:

```
RequirementProcessor.execute(
    this.logger,
    this.commandRequirements,
    requirementMessages.getMessage(
        'error:requirementCheckFailed'
    ),
    requirementMessages.getMessage(
        'error:requirementCheckFailed:recommendation'
    )
)
```

Since every command is going to pass the same values for the 3rd & 4th param, I think we can just have `RequirementProcessor.execute(this.logger, this.commandRequirements)` and have the `execute` method internally reference those 2 messages.

I also don't want to boil the whole ocean here so take the above as just a suggestion that you can skip for this PR.

RE the `logger` I'm ok with removing it too.

sfdctaka: Yes, you are right, we'd have to repeat, because we always use the same failure and recommendation messages regardless of which command the requirements were checked from.

If this will keep remaining to be the case then offering the caller to specify messages may be a moot thing to do. Instead, we can just keep those messages to be shown as specified within the `execute()` function.

I made the third and fourth parameters have default values in case the caller don't want to be concerned with specifying them but offering such wiggle room may have brought this conversation. Let me remove those parameters and also clean up the resource file.

And also remove `logger` ;)

maliroteh-sf (approved):
LGTM now 👍

src/common/Requirements.ts
```diff
@@ -103,73 +110,63 @@ export function WrappedPromise(
...
 
+export class RequirementProcessor {
     /**
-     * Executes all of the base and additional requirement steps.
-     * @returns A SetupTestResult containing all of the results of the requirement tests
+     * Executes all of the base and command requirement checks.
      */
-    public async executeSetup(): Promise<SetupTestResult> {
-        const testResult: SetupTestResult = {
+    public static async execute(requirements: {
```
khawkins, line 122: `requirements ` can be typed as `CommandRequirements`, yes?

src/common/Requirements.ts
```diff
@@ -103,73 +110,63 @@ export function WrappedPromise(
...
+        [key: string]: RequirementList;
+    }): Promise<void> {
+        const testResult: RequirementCheckResult = {
             hasMetAllRequirements: true,
             tests: []
         };
 
         let totalDuration = 0;
         let allRequirements: Requirement[] = [];
```
khawkins, line 131: I'd rename this to `enabledRequirements`. It's actually misleading now to be called `allRequirements`.

khawkins (approved):
LGTM!

PR #36, opened 2021-04-14
differ initialization to caller

maliroteh-sf:
Making it the responsibility of the caller to initialize a command. If we invoke a command from command line then SFDX will initialize the command already. If we instantiate a command object in code then we would be responsible for initializing before using the command.


sfdctaka (approved):
LGTM!

PR #40, opened 2021-05-20
Ground work for switching to LWR server

maliroteh-sf:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/lwc-dev-mobile-core",
   "description": "Core module supporting Salesforce CLI mobile extension plug-ins",
-  "version": "1.0.0",
+  "version": "1.0.0-beta12",
```
maliroteh-sf, line 4: should it be `1.1.0-beta1` instead? 🤔

src/common/AndroidLauncher.ts
```diff
@@ -102,8 +102,7 @@ export class AndroidLauncher {
                 const port = useServer ? serverPort : undefined;
 
                 if (PreviewUtils.isTargetingBrowser(targetApp)) {
-                    const compPath = PreviewUtils.prefixRouteIfNeeded(compName);
-                    const url = `${address}:${port}/lwc/preview/${compPath}`;
+                    const url = `${address}:${port}`;
```
maliroteh-sf, line 105: with LWR server, the root url previews the component.

src/common/Common.ts
```diff
@@ -91,39 +92,41 @@ export class CommandLineUtils {
...
 
     /**
      * Checks to see if a platform flag has a valid value.
      * @param platformFlag The platform flag.
-     * @returns True if flag is valid (i.e either targeting iOS or Android).
+     * @param includeDesktop Indicates whether Desktop is allowed as a target platform. Defaults to false.
+     * @returns True if flag is valid.
      */
-    public static platformFlagIsValid(platformFlag: string) {
+    public static platformFlagIsValid(
```
maliroteh-sf, line 118: The changes to this file are my attempt at using a simple way (with minimal coding) of adding support for having commands that might want to target Desktop as a platform. With the exception of  the Preview command, currently most of our commands don't need to target desktop as a platform and that is why `includeDesktop` defaults to `false`. When validating its parameters, the Preview command will call `platformFlagIsValid` and pass `includeDesktop=true`.

This is a simple way of having the Preview command handle all 3 platform, instead of having code in our vscode extension to detect Desktop as a special case and handle it by itself, but then let the Preview command handle Android/iOS platforms. With the simple changes in this file, the code for our vscode extension will be greatly simplified as it will now just pass the parameters for any platform to the Preview command and let it do its thing.

src/common/Common.ts
```diff
@@ -91,39 +92,41 @@ export class CommandLineUtils {
...
     }
 
-    /**
-     * Helper method for resolving flag values.
-     * @param flag The input flag.
-     * @param defaultValue The default value for a flag.
-     * @returns If the input flag can be cast to a string that is not undefined/null/empty then
-     * the string value will be returned. Otherwise, the provided default value will be returned.
-     */
-    public static resolveFlag(flag: any, defaultValue: string): string {
```
maliroteh-sf, line 118: This is not needed anymore.

src/common/PlatformConfig.ts
```diff
@@ -40,7 +40,6 @@ export class PlatformConfig {
 // tslint:disable-next-line: max-classes-per-file
 export class IOSConfig {
     public readonly minSupportedRuntime: string = '13';
-    public readonly defaultSimulatorName: string = 'SFDXSimulator';
```
maliroteh-sf, line 43: These are not needed anymore.

src/common/PreviewUtils.ts
```diff
@@ -55,18 +55,6 @@ export class PreviewUtils {
         );
     }
 
-    /**
-     * Checks to see if a component route needs to be prefixed.
-     * @param compName A component route or name.
-     * @returns The updated route which now starts with `c/` if compName did not start with `c/` already.
-     */
-    public static prefixRouteIfNeeded(compName: string): string {
```
maliroteh-sf, line 63: This is not needed anymore.

package.json
khawkins: Yes—or whatever x.y.z-beta1 we choose where x or y or z needs to increment past where we are today.

khawkins: I'm also not sure that we should go the linear route from what `main` is versioning, either. If we need to make another release (in any capacity) of the -core package before we promote this, we're going to get into confusing territory, version-wise.

I'd propose the following:

- Version this as `1.0.0-lwr1`. The suffix doesn't have to be strict to "beta".
- When we publish, publish this with the "lwr" tag, rather than the "beta" tag.

Given that this is a feature branch with an indeterminate longevity—rather than a mainline continuance—I would prefer we keep the versioning separate, between them.

src/common/Common.ts
khawkins: I think that's fine for the general purposes of writing a consolidation, as that's likely something that should be done in the longer term. We should work with the platform dev team to decide what a final home for that logic should look like, that can satisfy both projects. But I agree that at a minimum this is a good exercise to prove out how that should work.

khawkins (commented):
Everything looks good to me, but let's settle on our versioning strategy before merging. 👍

package.json
maliroteh-sf: oh I didn't know `npm` allows us to put anything we want after the version. I thought `beta` was a special case. In that case `1.0.0-lwr` is the best option.

PR #48, opened 2021-12-10
upgrade package versions

maliroteh-sf:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/lwc-dev-mobile-core",
   "description": "Core module supporting Salesforce CLI mobile extension plug-ins",
-  "version": "1.0.0",
+  "version": "1.1.0",
```
maliroteh-sf, line 4: Should this be 1.0.1 or 2.0.0 instead, or 1.1.0 is good enough?

khawkins (approved):
My gut feels good with 1.1.0. We're not making any API changes, but are somewhat significantly updating our underlying dependencies. I could find 1.0.1 in my heart too. No need to bump to a major version, IMO.

PR #49, opened 2021-12-13
switch to using eslint

maliroteh-sf:


src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { FlagsConfig, SfdxCommand } from '@salesforce/command';
-import { Logger, Messages, SfdxError } from '@salesforce/core';
```
sfdctaka, line 8: I guess `SfdxError` was not used?

src/common/AndroidUtils.ts
```diff
@@ -1377,9 +1377,9 @@ export class AndroidUtils {
         return `'${pathName}'`;
     }
 
-    private static async fetchInstalledSystemImages(
-        androidApi: string
```
sfdctaka, line 1381: Nice cleanup.

src/cli/commands/force/lightning/local/setup.ts
maliroteh-sf: Yes, and switching over to eslint exposed this unused import.

src/common/PreviewUtils.ts
```diff
@@ -89,9 +89,9 @@ export class PreviewUtils {
                 errorMessage: errorText,
                 passed: isValid
             });
-        } catch (err: any) {
+        } catch (err) {
             return Promise.resolve({
-                errorMessage: err,
+                errorMessage: (err as any).toString(),
```
sfdctaka, line 94: Just curious. If we left `err` as it is instead of `(err as any).toString()` by virtue of `errorMessage` being a string would we get automatic conversion and not have to do casting to `any` and then getting the string by calling `toString()`?

src/common/__tests__/AndroidUtils.test.ts
```diff
@@ -197,28 +197,28 @@ describe('Android utils', () => {
...
     });
 
     test('Should utilize cache for subsequent calls', async () => {
         jest.spyOn(CommonUtils, 'executeCommandAsync').mockImplementation(
             myCommandBlockMock
         );
-        let packages = await AndroidUtils.fetchInstalledPackages();
-        packages = await AndroidUtils.fetchInstalledPackages();
-        packages = await AndroidUtils.fetchInstalledPackages();
+        await AndroidUtils.fetchInstalledPackages();
```
sfdctaka, line 208: Nice cleanups.

src/common/PreviewUtils.ts
maliroteh-sf: That wouldn't work and ESLint would still throw an error. This is b/c in the newer version of TypeScript, the `error` in a try-catch block is explicitly marked as `unknown` type so it `must` be cast to something before it can be used.

sfdctaka (approved):
LGTM!

PR #51, opened 2022-03-07
@W-9812000 - Install android command line tools

pvandyk:


src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -67,16 +77,37 @@ export class Setup extends SfdxCommand implements HasRequirements {
...
-                this.flags.platform
-            )
-                ? new AndroidEnvironmentRequirements(
-                      this.logger,
-                      this.flags.apilevel
-                  )
-                : new IOSEnvironmentRequirements(this.logger);
-
-            this._commandRequirements.setup = requirements;
+            if (CommandLineUtils.platformFlagIsAndroid(this.flags.platform)) {
```
maliroteh-sf, line 80: hmm... not sure if I like this approach. `setup.ts` was supposed to be the place where we look at the selected platform then call upon the platform specific environment requirements to be checked (i.e `AndroidEnvironmentRequirements` or `IOSEnvironmentRequirements`) and abstract away the list of requirements per platform.

Maybe I'm missing something but based on the changes in this file, I don't see why we need to move away from that design.

src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -44,7 +51,10 @@ export class Setup extends SfdxCommand implements HasRequirements {
 
     public async run(): Promise<any> {
         this.logger.info(`Setup command called for ${this.flags.platform}`);
-        return RequirementProcessor.execute(this.commandRequirements);
+        return RequirementProcessor.execute(
+            this.commandRequirements,
+            `${this.flags.platform}`
```
maliroteh-sf, line 56: 👍

src/common/AndroidEnvironmentRequirements.ts
```diff
@@ -16,27 +17,76 @@ const messages = Messages.loadMessages(
...
-            new EmulatorImagesRequirement(logger, apiLevel)
+            new AndroidSDKInstallTask(logger)
+        ];
+    }
+}
+
+export class Java8AvailableRequirements implements RequirementList {
+    public requirements: Requirement[] = [];
+    public enabled = true;
+    public title = 'Java 8 Requirements';
```
maliroteh-sf, line 35: If these strings are user facing then they should be coming from the message file

src/common/AndroidEnvironmentRequirements.ts
```diff
@@ -16,27 +17,76 @@ const messages = Messages.loadMessages(
...
+    constructor(logger: Logger) {
+        this.requirements = [new Java8AvailableRequirement(logger)];
```
maliroteh-sf, line 37: Why did you decide to create these extra classes as `RequirementList` but with only 1 requirement in the list? Couldn't we just stick to the original class extending `Requirement`?

src/common/AndroidEnvironmentRequirements.ts
```diff
@@ -65,9 +116,44 @@ export class AndroidSDKRootSetRequirement implements Requirement {
                 )
             );
         } else {
-            return Promise.reject(new SfdxError(this.unfulfilledMessage));
+            return Promise.resolve(this.unfulfilledMessage);
```
maliroteh-sf, line 119: Add a comment about why we're still resolving instead of rejecting

src/common/Requirements.ts
```diff
@@ -182,6 +225,9 @@ export class RequirementProcessor {
                                         ? Promise.resolve()
                                         : Promise.reject(new Error());
                                 }),
+                            skip: () =>
+                                requirement.skipped == true &&
+                                requirement.title,
```
maliroteh-sf, line 230: Not sure if I understand the logic here... why do we need the `&& requirement.title`?

src/common/Requirements.ts
```diff
@@ -120,38 +122,79 @@ export class RequirementProcessor {
...
-        const requirementTasks = new Listr(
+        return new Listr(
             {
                 task: (_rootCtx, rootTask): Listr => {
                     const subTasks = new Listr([], {
                         concurrent: true,
                         exitOnError: false
                     });
-                    for (const requirement of enabledRequirements) {
+                    for (const requirement of requirementList.requirements) {
```
maliroteh-sf, line 197: If I understand your changes correctly, we used to have an array of requirements and now we have an array of array of requirements, right? If so then you need to update your logic in lines 203-206 to (a) make the current array as `hasMetAllRequirements = false` and also mark the root array (containing all other arrays) with `hasMetAllRequirements = false`

src/cli/commands/force/lightning/local/setup.ts
pvandyk: The issue was that the current model is a single task with all 6 subtasks.  To add optional install tasks for all tasks (if we want to eventually resolve all tasks with installs), it would mean we would have 1 task with 12 subtasks.  Of which 6 would be skipped. You'd have to look carefully at the description to figure out which task was which.  Instead, I thought it made sense to make each "component" as separate task with two subtasks.  One to check and one to install.  So you can see here, you can clearly see each task.  And the success and status of each subtask.

Screen Shot 2022-03-03 at 3 40 32 PM
.

pvandyk: The original list showed "Requirements" with a long list of requirements.  Not really breaking things into the sub-components.  For example.  These are just the component checks, without the install.  WIth installs, there would be another 6 under the main "Requirements" heading.
Screen Shot 2022-03-03 at 3 51 46 PM

src/common/AndroidEnvironmentRequirements.ts
pvandyk: I could consider adding that to this PR.  But thought I already let scope creep enough.  I can add a ticket for followup afterwards.

src/common/AndroidEnvironmentRequirements.ts
pvandyk: This is the anticipation of creating the install tasks for each in the future similar to the SDK root task.

src/common/Requirements.ts
pvandyk: The logic to resolve requirements will resolve each main task separately.  It waits for each task (including all subtasks) to resolve before checking the next task.  And will check the array of results before proceeding.  See lines 149-152.

src/cli/commands/force/lightning/local/setup.ts
maliroteh-sf: But you could still move the lines here for setting up r1-r5 to `AndroidEnvironmentRequirements` and still achieve the same thing, right? That way you don't need to modify this file at all.

src/common/Requirements.ts
pvandyk: Yup.  Removed.  Thanks!

src/common/AndroidEnvironmentRequirements.ts
maliroteh-sf: Adding them to the message file is really small work. You can just copy-paste them there and then pull them here. It would be nice to include that as part of this PR for a complete work.

src/cli/commands/force/lightning/local/setup.ts
pvandyk: ok, I'd have to modify the IOSEnvironmentRequirements file to return a single command key, but yes, I think that would make sense to pull this logic into Android and have setup agnostic of the tasks for each platform..

src/common/AndroidEnvironmentRequirements.ts
```diff
@@ -65,29 +164,68 @@ export class AndroidSDKRootSetRequirement implements Requirement {
                 )
             );
         } else {
-            return Promise.reject(new SfdxError(this.unfulfilledMessage));
+            // We will resolve rather than reject since we indicate that the SDK was not found.
```
maliroteh-sf, line 167: 👍

src/common/AndroidEnvironmentRequirements.ts
```diff
@@ -8,35 +8,133 @@ import { Logger, Messages, SfdxError } from '@salesforce/core';
...
     '@salesforce/lwc-dev-mobile-core',
     'requirement-android'
 );
 
-export class AndroidEnvironmentRequirements implements RequirementList {
+export class AndroidEnvironmentRequirements {
+    public commandRequirements: CommandRequirements;
+
+    constructor(logger: Logger, apiLevel?: string) {
+        this.commandRequirements = {};
```
maliroteh-sf, line 28: 👍

src/common/IOSEnvironmentRequirements.ts
```diff
@@ -9,17 +9,32 @@ import util from 'util';
...
 );
 
-export class IOSEnvironmentRequirements implements RequirementList {
+export class IOSEnvironmentRequirements {
+    public commandRequirements: CommandRequirements;
+
+    constructor(logger: Logger) {
+        this.commandRequirements = {};
+
+        this.commandRequirements.r1 = new IOSRequirements(logger);
```
maliroteh-sf, line 30: Doesn't have to be done in this PR if you don't want scope creep, but it would be nice to break this up similar to the Android counterpart.

src/common/AndroidEnvReqResolver.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    private async installAndroidCommandLineTools(): Promise<
+        string | undefined
+    > {
+        // download command line tools
+
+        // unzip and move to the latest folder
+        const unzipCommand = ''.concat(
+            'unzip /Users/pvandyk/Downloads/commandlinetools-mac-8092744_latest.zip -d $ANDROID_HOME/cmdline-tools/latest',
```
dbreese, line 61: `/Users/pvandyk`?

dbreese: Also, not sure I understand what this is doing -- appears it is extracting the zip into $ANDROID_HOME/cmdline-tools/latest, then moving all files from cmdline-tools/latest/cmdline-tools/* into cmdline-tools/latest? 

If so, then perhaps unzip to a tmp folder and just move it to where it belongs?

```
unzip somefile.zip -d $ANDROID_HOME/tmp-cmdline-tools && \
    mkdir -p $ANDROID_HOME/cmdline-tools/latest && \
    mv $ANDROID_HOME/tmp-cmdline-tools $ANDROID_HOME/cmdline-tools/latest
```

src/common/AndroidEnvReqResolver.ts
```diff
@@ -48,13 +51,45 @@ export class AndroidCommandLineToolsResolver implements RequirementResolver {
...
+            ' && ',
+            'mv $ANDROID_HOME/cmdline-tools/latest/cmdline-tools/* $ANDROID_HOME/cmdline-tools/latest',
+            ' && ',
+            'rm -rf $ANDROID_HOME/cmdline-tools/latest/cmdline-tools'
+        );
+        CommonUtils.executeCommandSync(unzipCommand);
+
+        // this is needed because unzipping doesn't register the sdkmanager
+        // it will install to latest-2, so remove and install
+        const installCommand = ''.concat(
```
maliroteh-sf, line 71: @pvandyk changes that go into this repo should work on mac/unix/linux and on windows platforms (take a look [here](https://github.com/forcedotcom/lwc-dev-mobile-core/blob/6edd4c22a22517d52f46616ff49a857978a05e6d/src/common/AndroidUtils.ts#L840-L842) for an example). I don't think these commands work on Windows as they are right now.

src/common/AndroidEnvReqResolver.ts
pvandyk: I maybe able to unzip the folder directly into latest.  I'll look into that, depending on the unzip node solution we choose.

PR #52, opened 2022-03-11
@ W-10055362  - select unique iOS simulator

pvandyk:


khawkins (approved):
Overall, looks great! A couple of comments:

1. It wouldn't hurt to have a unit test around this, in `IOSUtils.test.ts`. We do mock/overload the `getSuportedSimulators()` output (see [`IOSMockData.mockRuntimeDevices`](https://github.com/forcedotcom/lwc-dev-mobile-core/blob/8f4a0ed5ba3a4e96dc9fbb769804b88d3bffbf1e/src/common/__tests__/IOSMockData.ts#L105-L179), and [one of the `getSupportedSimulators()` tests](https://github.com/forcedotcom/lwc-dev-mobile-core/blob/8f4a0ed5ba3a4e96dc9fbb769804b88d3bffbf1e/src/common/__tests__/IOSUtils.test.ts#L344-L353) as an example), so we could e.g. make sure `udid` searches are responsive.
2. You'll want to bump the `version` value in `package.json`. I'd be supportive of `1.2.1`, since we're really just enhancing the behavior of an existing API. But `1.3.0` would be okay too.

Otherwise, LGTM. 👍

src/common/IOSTypes.ts
```diff
@@ -80,9 +91,9 @@ export class IOSSimulatorDevice {
     }
 
     /**
-     * A string representation of an IOSSimulatorDevice which includes Name, Runtime Id
+     * A string representation of an IOSSimulatorDevice which includes Name, Runtime Id, and udid
      */
     public toString(): string {
-        return `${this.name}, ${this.runtimeId}`;
+        return `${this.name}, ${this.runtimeId}, ${this.udid}`;
```
maliroteh-sf, line 97: If I recall correctly we didn't include `udid` originally b/c we didn't think that would be useful info for the user. Often times the user cares more about human-readable info, so we kept name + runtime info but not `udid`. Are you including it here b/c you want to use it in another place (for example by parsing the output)?

pvandyk: The idea @maliroteh-sf is that you will be able to list iOS devices with
sfdx force:lightning:local:device:list -p iOS

And then to start an iOS device and start the exact device you want, you can use:
sfdx force:lightning:local:device:start -p iOS -t "<udid>"

You can still use name:
sfdx force:lightning:local:device:start -p iOS -t "<name>"

But if you do, we will select the first one in the list since the names are not unique for a device with multiple simulators for each version of iOS.  By displaying the udid in the list, you will now be able to choose the specific iOS device to start and therefore be able to start the device with a particular ios version.

pvandyk:
> Overall, looks great! A couple of comments:
> 
> 1. It wouldn't hurt to have a unit test around this, in `IOSUtils.test.ts`. We do mock/overload the `getSuportedSimulators()` output (see [`IOSMockData.mockRuntimeDevices`](https://github.com/forcedotcom/lwc-dev-mobile-core/blob/8f4a0ed5ba3a4e96dc9fbb769804b88d3bffbf1e/src/common/__tests__/IOSMockData.ts#L105-L179), and [one of the `getSupportedSimulators()` tests](https://github.com/forcedotcom/lwc-dev-mobile-core/blob/8f4a0ed5ba3a4e96dc9fbb769804b88d3bffbf1e/src/common/__tests__/IOSUtils.test.ts#L344-L353) as an example), so we could e.g. make sure `udid` searches are responsive.
> 2. You'll want to bump the `version` value in `package.json`. I'd be supportive of `1.2.1`, since we're really just enhancing the behavior of an existing API. But `1.3.0` would be okay too.
> 
> Otherwise, LGTM. 👍

Thanks Kevin, I created the test.  And updated the version to `1.3.0` since we're adding another feature and not fixing a bug.

PR #55, opened 2022-06-20
Handle Android codename package versions gracefully

khawkins:
This should handle the scenario where a user has Android codename version packages in their local SDK. Before, these would result in an unhandled error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/lwc-dev-mobile-core",
   "description": "Core module supporting Salesforce CLI mobile extension plug-ins",
-  "version": "1.3.0",
+  "version": "1.3.1",
   "author": {
     "name": "Meisam Seyed Aliroteh",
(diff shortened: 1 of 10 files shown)


maliroteh-sf (approved):
LGTM 👍

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/lwc-dev-mobile-core",
   "description": "Core module supporting Salesforce CLI mobile extension plug-ins",
-  "version": "1.3.0",
+  "version": "2.0.0",
```
khawkins, line 4: @maliroteh-sf I realized late in the game that technically I'm making public API changes here. I opted to bump the major version, since I think it's really only our packages that depend on this one. But I could also tack back to an error-raising approach consistent with the previous API version, and save the API changes for a later date.

What do you think?

maliroteh-sf: I think increasing the major version is more inline with semantic versioning, since you're updating the API.

PR #57, opened 2022-11-14
Update dependencies

maliroteh-sf:
- Update dependencies
- `SfdxError` is deprecated and should be changed to `SfError`
- Minor code clean up


package.json
```diff
@@ -31,38 +31,38 @@
   ],
   "bugs": "https://github.com/forcedotcom/lwc-dev-mobile-core/issues",
   "dependencies": {
-    "@oclif/config": "^1.18.2",
-    "@salesforce/command": "^4.2.1",
-    "@salesforce/core": "^2.31.1",
-    "ajv": "^8.8.2",
+    "@oclif/config": "^1.18.6",
```
khawkins, line 34: I had an open question about `@oclif/config` specifically: their [repo](https://github.com/oclif/config) lists this package as deprecated, with the suggestion to move to [`@oclif/core`](https://github.com/oclif/core). Should we take this opportunity to do so? The `@oclif/core` repo has a [migration guide](https://github.com/oclif/core/blob/main/MIGRATION.md) we could use.

This package is only receiving security updates at this point, and not sure how long that will be the case.

maliroteh-sf: hmm... sure that's a good idea... let me take a look at that.

yarn.lock
```diff
@@ -2,356 +2,362 @@
...
-    "@babel/helper-compilation-targets" "^7.16.0"
-    "@babel/helper-module-transforms" "^7.16.0"
-    "@babel/helpers" "^7.16.0"
-    "@babel/parser" "^7.16.0"
-    "@babel/template" "^7.16.0"
-    "@babel/traverse" "^7.16.0"
-    "@babel/types" "^7.16.0"
+"@ampproject/remapping@^2.1.0":
+  version "2.2.0"
+  resolved "https://nexus-proxy-prd.soma.salesforce.com/nexus/content/groups/npm-all/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
```
khawkins, line 7: `yarn` updates need to run against the public registry (https://registry.yarnpkg.com).

package.json
```diff
@@ -31,42 +31,43 @@
...
-    "@oclif/dev-cli": "^1.26.9",
-    "@types/jest": "^26.0.20",
-    "@types/node": "^16.11.12",
-    "@typescript-eslint/eslint-plugin": "^5.7.0",
-    "@typescript-eslint/parser": "^5.7.0",
-    "eslint": "^8.4.1",
-    "eslint-config-prettier": "^8.3.0",
-    "eslint-plugin-import": "^2.25.3",
-    "eslint-plugin-jsdoc": "^37.2.0",
+    "@oclif/dev-cli": "^1.26.10",
```
khawkins, line 44: Looks like, with an upgrade to `@oclif/core` (or maybe independent of it), we don't need this package anymore?

yarn.lock
maliroteh-sf: very good catch... I forgot to update my environment setting to point to the public registry.

package.json
maliroteh-sf: Yes you're right, this is no longer needed.

khawkins (approved):
LGTM!

sfdctaka (approved):
LGTM!

PR #61, opened 2023-01-19
Add utility functions to create/read text files

maliroteh-sf:
Also upgrade dependencies.


src/common/CommonUtils.ts
```diff
@@ -369,5 +369,40 @@ export class CommonUtils {
...
+     */
+    public static async readTextFile(path: string): Promise<string> {
+        return new Promise((resolve, reject) => {
+            fs.readFile(path, function (err, data) {
+                if (err) {
+                    const msg = `Error reading file ${path} - ${err}`;
+                    CommonUtils.logger.error(msg);
+                    reject(err);
+                }
+                resolve((data ?? '').toString());
```
dbreese, line 402: Do we need to do anything in particular for encoding, or is converting to a plain string good enough for any text file?

dbreese (approved):
LGTM - just minor question about encoding?

src/common/CommonUtils.ts
maliroteh-sf: For now I think plain text is good enough. We can always update in the future if we need support for other encodings.

PR #64, opened 2023-02-14
Add ability to run command using spawn()

maliroteh-sf:
After reading [this](https://www.freecodecamp.org/news/node-js-child-processes-everything-you-need-to-know-e69498fe970a/) blog post, I'm adding a method for running commands using `child_process.spawn()`. According to that blog post, we should be using `spawn` instead of `exec` if we expect to have a long running process which might produce a large amount of output. By adding this new method, we can use it in `lwc-dev-mobile` when running WDIO related commands which may produce large amount of output.

Furthermore, when using `exec` the content of `stdout` and `stderr` will be captured and send to the caller **_after_** the execution is done. Whereas using `spawn` we have the option of streaming the stdio to the caller while the command is running.

### DIFF
--- a/src/common/CommonUtils.ts
+++ b/src/common/CommonUtils.ts
@@ -200,7 +200,7 @@ export class CommonUtils {
         command: string,
         stdioOptions: StdioOptions = ['ignore', 'pipe', 'ignore']
     ): string {
-        CommonUtils.logger.debug(`Executing command: '${command}'.`);
+        CommonUtils.logger.debug(`Executing command: '${command}'`);
         try {
             return childProcess
                 .execSync(command, {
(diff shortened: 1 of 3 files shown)


src/common/CommonUtils.ts
```diff
@@ -215,51 +215,99 @@ export class CommonUtils {
...
-                                msg = `${msg}\nstderr:\n${stderr}`;
-                            }
-                            if (stdout && stdout.length > 0) {
-                                msg = `${msg}\nstdout:\n${stdout}`;
-                            }
-
-                            CommonUtils.logger.error(msg);
-                            reject(error);
-                        } else {
-                            resolve({ stdout, stderr });
```
maliroteh-sf, line 251: Reverting these back to the way it was before. These were added in order to be used by our WDIO-related commands in `lwc-dev-mobile`, which now will use `spawnCommandAsync` instead.

sfdctaka (approved):
LGTM!

PR #65, opened 2023-02-23
Upgrade dependencies to address deprecations

maliroteh-sf:
Upgrade dependencies to the latest versions. Move away from the deprecated `@salesforce/command` package and migrate to `@salesforce/sf-plugins-core` instead.

### DIFF
--- a/src/cli/commands/force/lightning/local/setup.ts
+++ b/src/cli/commands/force/lightning/local/setup.ts
@@ -12,7 +12,10 @@ import {
     FlagsConfigType
 } from '../../../../../common/Common';
 import { IOSEnvironmentRequirements } from '../../../../../common/IOSEnvironmentRequirements';
-import { RequirementProcessor } from '../../../../../common/Requirements';
+import {
+    CommandRequirements,
+    RequirementProcessor
+} from '../../../../../common/Requirements';
 
 // Initialize Messages with the current plugin directory
 Messages.importMessagesDirectory(__dirname);
@@ -25,7 +28,7 @@ const messages = Messages.loadMessages(
 );
 
 export class Setup extends BaseCommand {
-    protected commandName = 'force:lightning:local:setup';
+    protected _commandName = 'force:lightning:local:setup';
 
     public static readonly description =
         messages.getMessage('commandDescription');
@@ -36,19 +39,11 @@ export class Setup extends BaseCommand {
     ];
 
     public static readonly flags = {
-        ...CommandLineUtils.createFlagConfig(
-            FlagsConfigType.ApiLevel,
-            false,
-            Setup.examples
-        ),
-        ...CommandLineUtils.createFlagConfig(
-            FlagsConfigType.Platform,
-            true,
-            Setup.examples
-        )
+        ...CommandLineUtils.createFlag(FlagsConfigType.ApiLevel, false),
+        ...CommandLineUtils.createFlag(FlagsConfigType.Platform, true)
     };
 
-    public async run(): Promise<any> {
+    public async run(): Promise<void> {
         this.logger.info(
             `Setup command called for ${this.flagValues.platform}`
         );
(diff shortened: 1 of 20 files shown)


src/common/__tests__/CommonUtils.test.ts
```diff
@@ -268,6 +268,38 @@ describe('CommonUtils', () => {
         ]);
     });
 
+    // Disabling this test for now. It runs & passes locally but it fails
```
maliroteh-sf, line 271: As part of this PR I added this new test to increase coverage but as I've stated in the comments, while it passes locally, it fails in CI. I decided to keep the code here instead of throwing it out hoping that in the future we might be able to figure out why this is happening.

src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -29,56 +27,41 @@ const messages = Messages.loadMessages(
...
+        ...CommandLineUtils.createFlag(FlagsConfigType.Platform, true)
+    };
 
-        return super
-            .init()
-            .then(() => Logger.child('force:lightning:local:setup', {}))
-            .then((logger) => {
-                this.logger = logger;
-                return LoggerSetup.initializePluginLoggers();
-            });
```
sfdctaka, line 64: Are we not responsible for setting up the logger anymore?

src/common/BaseCommand.ts
```diff
@@ -0,0 +1,81 @@
...
+            .then((parserOutput) => {
+                this._flagValues = parserOutput.flags;
+                return Logger.child(this.commandName);
+            })
+            .then((logger) => {
+                this._logger = logger;
+                return LoggerSetup.initializePluginLoggers();
+            })
+            .then(() => this.populateCommandRequirements());
+    }
```
sfdctaka, line 62: NVM earlier comment. It's here :)

maliroteh-sf: Yes 😀

I don't know why I didn't do it this way a long while ago when I initially added the commands in this repo. But now that I've introduced `BaseCommand`, the `init` block for all of our commands (from all repos) can be deleted since `BaseCommand` handles it... and this makes it much cleaner

src/common/BaseCommand.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    // Loops over all of the flags of a command and checks to see
+    // if a flag has a `validate` function defined. If so then we
+    // mark it to use `CommandLineUtils.flagParser` as a parser,
+    // which will execute the validation function.
+    private injectFlagParser() {
+        const flags = (<typeof SfCommand>this.constructor).flags;
+        const flagEntries = Object.entries(flags);
+        flagEntries.forEach((flag) => {
+            const configEntries = Object.entries(flag[1]);
```
sfdctaka, line 72: What is `flag[1]`? The value associated with the flag?

maliroteh-sf: `Object.entries` returns an array of KeyValue pairs where item[0] is key and item[1] is value. In this context `flag[1]` holds another object which defines the configs for that flag. By default, `oclif` flag configs contain a parser that just accepts whatever value is passed in for that flag. Here in this function we check to see if `validate` is added to the configs for a flag. If so then we override the default parser with our parser which runs the validation check and only accepts the flag value if validation passes.

sfdctaka: Do you think it would be helpful if we put the above in comment?

maliroteh-sf: You mean in addition to the comment that I've already put on the method? I guess I can reword that comment to include more info.

sfdctaka: It would be a nice addition, I think, although a nit, yes. I don't anticipate working with flags is going to be the dominant work in lwc-dev-mobile-core in the future but if anyone needs to visit and find what's going on right away a little note should be beneficial.

sfdctaka (approved):
Pulled your code and tested locally. LGTM!

src/common/BaseCommand.ts
```diff
@@ -0,0 +1,89 @@
...
+        const flagEntries = Object.entries(flags);
+        flagEntries.forEach((flag) => {
+            // Object.entries returns an array of KeyValue pairs where item[0]
+            // is key and item[1] is value. In this context flag[1] holds another
+            // object which defines the configs for that flag. oclif adds a default
+            // parser to flag configs for all flags. This parser just accepts any value
+            // that is passed in for that flag without validating anything.
+            // Here we check to see if validate is added to the configs for a flag.
+            // If so then we override the default parser with our parser which runs
+            // the validation check and only accepts the flag value if validation passes.
```
sfdctaka, line 79: 🎉

PR #66, opened 2023-02-24
Add support for logLevel and json flags

maliroteh-sf:
Bring back these flags for backwards compatibility since SFDX VSCode Extension is still using older version of `SfCommand`

### DIFF
--- a/src/common/BaseCommand.ts
+++ b/src/common/BaseCommand.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { SfCommand } from '@salesforce/sf-plugins-core';
-import { Logger } from '@salesforce/core';
+import { Logger, LoggerLevel } from '@salesforce/core';
 import { CommandLineUtils } from './Common';
 import { LoggerSetup } from './LoggerSetup';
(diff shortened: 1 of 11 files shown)


src/common/BaseCommand.ts
```diff
@@ -55,8 +55,18 @@ export abstract class BaseCommand
                 return Logger.child(this.commandName);
             })
             .then((logger) => {
+                // extract the log level flag (if any) and
+                // set the logger's level to this value
+                const logLevelFlag: string | undefined = (
```
maliroteh-sf, line 60: We were never actually using the value passed in for the `loglevel` flag (even in the past). Now that I'm bringing in these flags for backwards compatibility with SFDX VSCode plugin, I'm going to go ahead and actually wire it up too.

sfdctaka: Can you help me remember why we took the out the flags please?

maliroteh-sf: We didn't take out the flags. The older versions of `SfCommand` used to have these built-in but the new versions don't have these built-in anymore (not sure why though, and that's a different discussion).

It is now the implementors job to add whatever flags they want. I learned this when I was trying to launch commands from VSCode Extension using our latest builds of `lwc-dev-*` and saw errors on undefined `--json` and `--loglevel` flags.

PR #67, opened 2023-02-28
start cli action only when stdout supports TTY

maliroteh-sf:


src/common/CommonUtils.ts
```diff
@@ -81,7 +81,11 @@ export class CommonUtils {
      * @param status Optional status message for the action.
      */
     public static startCliAction(action: string, status?: string) {
-        ux.action.start(action, status, { stdout: true });
+        if (process.stdout?.isTTY === true) {
+            ux.action.start(action, status, { stdout: true });
+        } else {
+            console.log(`${action}... ${status ?? ''}`);
```
khawkins, line 87: Can you describe approximately what the UX difference will be here, for unattended execution of these tasks vs. user-initiated execution?

khawkins: And is that for the use case of something like CI, or would it also happen in situations like running the CLI commands from VSCode?

maliroteh-sf: I've described the issue in its corresponding work item.

Most of our SFDX commands show animated text-based spinners while the command is in progress. Essentially this is equivalent to `console.log(`${action}... ${status ?? ''}`)` + an animated spinner appended to it.

The issue with this is that, on some platforms like Windows or depending on how the sfdx process is launched (i.e from terminal vs vscode extension) it may be the case that the stdout for the process may not support TTY. This means that instead of an animated text-based spinner being displayed, the animated spinner would print a new line for each frame of the animation.

This PR checks to see if the stdout of the current process supports TTY or not and if so then we proceed as before but if not then we only do the `console.log` without the animated spinner that causes the issue.

maliroteh-sf: Here's a screenshot illustrating the issue when we run the SFDX LWC Preview command in VSCode extension:
Screenshot 2023-02-28 at 9 35 37 AM

maliroteh-sf: With the changes in this PR, the output would then be:
```
.
.
.
Start Preview... searching for device Pixel_5_API_30
Start Preview... searching for device Pixel_5_API_30
Start Preview... navigating mobile browser to http://10.0.2.2:3333/lwc/preview/c/testing
Launch Browser... Opening browser with url http://10.0.2.2:3333/lwc/preview/c/testing
Launch Browser... done
.
.
.
```

khawkins (approved):

PR #68, opened 2023-03-03
Add support for codenamed version strings

maliroteh-sf:
Currently our commands (Device List, Device Create, Setup, ...) only work when targeting Android versions that follow proper semantic versioning. However when beta versions of Android are available they are often versioned as a codename (Tiramisu, UpsideDownCake, ...) which are strings not following proper semantic versioning.

This PR adds support for codename versions to our tools.


sfdctaka (approved):
LGTM!

PR #75, opened 2023-08-28
Add support for desktop as a platform

maliroteh-sf:
Previously, plugin commands that are built on top of `lwc-dev-mobile-core` would only support `ios` or `android` as platforms. In this PR we add the ability for a plugin command to specify `desktop` as a platform as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/lwc-dev-mobile-core",
   "description": "Core module supporting Salesforce CLI mobile extension plug-ins",
-  "version": "3.2.0",
+  "version": "3.2.1",
   "author": {
     "name": "Meisam Seyed Aliroteh",
     "email": "maliroteh@salesforce.com",
--- a/src/common/Common.ts
+++ b/src/common/Common.ts
@@ -78,6 +78,7 @@ export enum FlagsConfigType {
 export class CommandLineUtils {
     public static IOS_FLAG = 'ios';
     public static ANDROID_FLAG = 'android';
+    public static DESKTOP_FLAG = 'desktop';
 
     /**
      * Checks to see if a flag is targeting iOS.
@@ -85,10 +86,7 @@ export class CommandLineUtils {
      * @returns True if flag is targeting iOS.
      */
     public static platformFlagIsIOS(input: string): boolean {
-        if (input) {
-            return input.toLowerCase() === CommandLineUtils.IOS_FLAG;
-        }
-        return false;
+        return (input ?? '').trim().toLowerCase() === CommandLineUtils.IOS_FLAG;
     }
 
     /**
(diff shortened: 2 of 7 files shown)


dbreese (approved):
LGTM

sfdctaka (approved):
LGTM!

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/lwc-dev-mobile-core",
   "description": "Core module supporting Salesforce CLI mobile extension plug-ins",
-  "version": "3.2.0",
+  "version": "3.2.1",
```
khawkins, line 4: Typically, SemVer dictates that when you're adding new public APIs, that necessitates a minor version bump.

maliroteh-sf: I have another PR coming in... I'll set the version to `3.3.0` in that PR

src/common/Common.ts
```diff
@@ -131,7 +127,29 @@ export class CommandLineUtils {
...
+            CommandLineUtils.platformFlagIsAndroid(platformFlag) ||
+            (includeDesktop &&
+                CommandLineUtils.platformFlagIsDesktop(platformFlag))
+        );
+    }
+
+    public static createFlag(
+        type: FlagsConfigType,
+        isRequired: boolean,
+        supportsDesktop = false
```
khawkins, line 151: I don't have great visibility on how plugins are going to use this, but in the absence of that visibility it feels a lot like pushing plugin-specific considerations down into the core library.

What would be a typical use case for one of our consuming plugins, here?

maliroteh-sf: In `lwc-dev-mobile` we define a number of plugin commands. Each command has a bunch of flags and there are a few flags (such as ApiLevel, Platform, LogLevel) that are  common to all commands. So the logic around common flags had been placed in `lwc-dev-mobile-core` and I'm just building on that.

As an example, here's what the `flags` definition for an updated `Preview` command would look like:
```
    public static readonly flags = {
        ...CommandLineUtils.createFlag(FlagsConfigType.ApiLevel, false),
        ...CommandLineUtils.createFlag(FlagsConfigType.LogLevel, false),
        ...CommandLineUtils.createFlag(FlagsConfigType.Platform, true, true),
        componentname: Flags.string({
            char: 'n',
            description: messages.getMessage('componentnameFlagDescription'),
            required: true
        }),
        configfile: Flags.string({
            char: 'f',
            description: messages.getMessage('configFileFlagDescription'),
            required: false,
            default: ''
        }),
        projectdir: Flags.string({
            char: 'd',
            description: messages.getMessage('projectDirFlagDescription'),
            required: false,
            default: process.cwd()
        }),
        .
        .
        .
    };
```

khawkins: Ah, interesting. Since there is just a handful or arguments and types associated with this functionality, I'm inclined to go with the change as coded, then. Pedantic overthinking follows below. 😉

The thing that is getting me is that we're effectively bifurcating the input parameter to `createFlag()` needed to configure the same input value: in this case, `type` and `supportsDesktop`. In a world where we needed to scale out this method to any number of supported flags, ideally we'd provide an input parameter data structure where each flag can specify further constraints. E.g.:

```typescript
CommandLineUtils.createFlag(
    {
        type: FlagsConfigType.Platform,
        supportedPlatforms: (PlatformTypeIOS | PlatformTypeAndroid | PlatformTypeDesktop
    },
    true
)
```

Or similar to Node.js methods (like `fs.readFile()`) that provide blanket "options" parameters:

```typescript
CommandLineUtils.createFlag(
    FlagsConfigType.Platform,
    true,
    {
        supportedPlatforms: PlatformTypeIOS | PlatformTypeAndroid | PlatformTypeDesktop
    }
)
```

Anyway, food for thought if we find ourselves getting unwieldy in our input management here.

maliroteh-sf: I could also update our enum definition

From
```
export enum FlagsConfigType {
    Platform,
    ApiLevel,
    LogLevel,
    Json
}
```

To
```
export enum FlagsConfigType {
    MobileOnlyPlatforms,
    MobileAndDesktopPlatforms,
    ApiLevel,
    LogLevel,
    Json
}
```

Then we could just call `CommandLineUtils.createFlag(FlagsConfigType.MobileAndDesktopPlatforms, ...)`. But this type of change (or the one you suggested) would require me to touch `ALL` of the commands and update them, whereas the approach I have added here in this PR uses a param with default value so as to not touch many of the commands and only touch the code for the `Preview` command.

PR #76, opened 2023-08-28
update preview URL format

maliroteh-sf:


dbreese (approved):

src/common/PreviewUtils.ts
```diff
@@ -55,18 +55,6 @@ export class PreviewUtils {
...
-     * @param compName A component route or name.
-     * @returns The updated route which now starts with `c/` if compName did not start with `c/` already.
-     */
-    public static prefixRouteIfNeeded(compName: string): string {
-        if (compName.toLowerCase().startsWith('c/')) {
-            return compName;
-        }
-        return 'c/' + compName;
-    }
-
```
sfdctaka, line 69: Wouldn't this be a breaking change if we remove it in the main branch? What happens to the downstream module that's using it? Or do we expect it to change as well if consuming this version of lwc-dev-mobile-core?

maliroteh-sf: We're going to take this into `browser-container` branch instead of `main` for now.

sfdctaka: Ah, didn't see that 👍

sfdctaka (approved):
LGTM!

PR #79, opened 2023-11-01
Clean up logger code

maliroteh-sf:


src/common/AndroidTypes.ts
```diff
@@ -17,14 +17,8 @@ export class AndroidPackages {
     /**
      * Initialize the logger used by AndroidPackages.
      */
-    public static async initializeLogger(
-        level?: LoggerLevelValue
-    ): Promise<void> {
```
maliroteh-sf, line 22: These were all remnant of old code. A long time ago we used to have each library create a logger (using `Logger.child`) whereas later on we moved away from that. Each library has a static instance of a logger and here we only need to set the level for the logger rather than create yet another child. I clean this up in all util classes.

src/common/BaseCommand.ts
```diff
@@ -52,23 +52,17 @@ export abstract class BaseCommand
...
             })
             .then((logger) => {
                 // extract the log level flag (if any) and
                 // set the logger's level to this value
-                const logLevelFlag: string | undefined = (
-                    this._flagValues.loglevel as string
-                )
-                    ?.trim()
-                    ?.toUpperCase();
-                const logLevel =
```
maliroteh-sf, line 65: This was doing the wrong thing. Casting as `(<any>LoggerLevel)[logLevelFlag]` was returning a `string` value instead of a `LoggerLevel` value so then the call to `logger.setLevel(logLevel);` which expects a `LoggerLevel` value would not accept the input param and instead would choose a default log level no matter what the actual value of `logLevel` was. Fixing this here using a helper function to get a `LogLevel` representation of the value.

dbreese (approved):
LGTM

sfdctaka (approved):
LGTM!

PR #83, opened 2024-05-13
upgrade dependencies and project structure

maliroteh-sf:
Followed the instruction [here](https://github.com/salesforcecli/cli/wiki/Code-Your-Plugin#generate-stuff) to upgrade the plugin project structure and below are some of the highlights of the changes that were necessary as a result of this upgrade:

- Upgraded dependencies to the latest
- Moved `module` from `commonjs` to `node16`
- Migrated tests from `Jest` to `Mocha` and now have the ability to add NUTs (if we decide to in the future)
- Moved test files from `__tests__` to a separate `test` folder to better organize `unit` and `NUT` tests
- Converted messages from `JSON` to `MD` format
- Now using `wireit` + `nyc` + `mocha` in our scripts 
- Now using `husky` + `lintstaged` hooks to run certain scripts to ensure code quality before committing to pushing changes to repo


src/cli/commands/force/lightning/local/__tests__/setup.test.ts
```diff
@@ -1,112 +0,0 @@
-/*
```
maliroteh-sf, line 1: All of the tests files where moved to a different path but for some reason, during diff, they are being shown as deleted + new ones added which makes code reviewing more difficult 😞

sfdctaka:
Seems like something makes `import` and `export` needing to have modules specified to end with `.js`. What is causing that? Is that the part of `commonjs` to `node16`?

maliroteh-sf:
> Seems like something makes `import` and `export` needing to have modules specified to end with `.js`. What is causing that? Is that the part of `commonjs` to `node16`?

Yes it is the result of moving to `node16`. Without it we would see this error
```
Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'.
```

.vscode/settings.json
```diff
@@ -0,0 +1,16 @@
...
+    "**/CVS": true,
+    "**/.DS_Store": true
+  },
+  "search.exclude": {
+    "**/dist": true,
+    "**/bin": true
+  },
+  "editor.tabSize": 4,
+  "editor.formatOnSave": true,
+  "rewrap.wrappingColumn": 120
```
sfdctaka, line 15: Are these something that we want to enforce in eslint?

.vscode/settings.json
```diff
@@ -0,0 +1,16 @@
...
+    "**/.git": true,
+    "**/.svn": true,
+    "**/.hg": true,
+    "**/CVS": true,
+    "**/.DS_Store": true
+  },
+  "search.exclude": {
+    "**/dist": true,
+    "**/bin": true
+  },
```
sfdctaka, line 12: My guess is these were generated from sf plugin generator and we're being a good citizen to follow the practice? I guess what I'm trying to determine is if we need settings.json at all?

src/common/BaseCommand.ts
```diff
@@ -82,10 +87,9 @@ export abstract class BaseCommand
             // If so then we override the default parser with our parser which runs
             // the validation check and only accepts the flag value if validation passes.
             const configEntries = Object.entries(flag[1]);
-            const hasValidate = configEntries.find(
-                (keyValuePair) => keyValuePair[0] === 'validate'
-            );
+            const hasValidate = configEntries.find((keyValuePair) => keyValuePair[0] === 'validate');
             if (hasValidate) {
+                // eslint-disable-next-line @typescript-eslint/unbound-method, no-param-reassign
```
sfdctaka, line 92: nit: looks like one-off one-liner. not sure if you would prefer every eslint disable up at the top of the file.

.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
sfdctaka, line 8: Should we have these in eslint?

sfdctaka (approved):
A nit and a few questions. Otherwise, looks good to me!

.vscode/settings.json
maliroteh-sf: I decided to leave things as generated after following steps [here](https://github.com/salesforcecli/cli/wiki/Code-Your-Plugin#generate-stuff), which means that some of these settings are not placed in eslint rules.

In my experience, if we move these to eslint, then some of them interfere with other styling rules (specially with prettier). For example if I add the eslint rule `'indent': ['error', 4]` then there will be cases where eslint will generate error b/c it thinks a line should start with 8 spaces but prettier wants the line to start with 12 spaces (I think Prettier style should actually be the one that wins b/c it results in a nicer style). Same issue goes with `'quotes': ['error', 'single'],` eslint rule where eslint would want to enforce single quote all the time but prettier would prefer you to use double quotes some time, so the two would clash.

As such I think we should just leave the rules as they are right now

.vscode/settings.json
maliroteh-sf: Yes they were generated by the sf plugin generator.... and yes I think we should keep settings.json as it has useful settings for vscode (specially the formatOnSave, where it would auto-format as soon as you save a file)

src/common/BaseCommand.ts
maliroteh-sf: Yes this is intentional. I actually prefer one-off disables over blanket file-level disables. The reason I left the file-level disables in other places is b/c there were several instances of those rule violations per file. As I mentioned [here](https://github.com/forcedotcom/lwc-dev-mobile-core/pull/83/files#r1598781315), going from `commonjs` to `node16` and also extending our tsconfig & eslint rules from `sf*` rules means stricter rule enforcement. Rather than boiling the whole ocean in this PR (which is already very large) I've opted for disabling some of these rules per file. We can always revisit in another PR and try to enable as many of them as we can.

.editorconfig
maliroteh-sf: see my comment [here](https://github.com/forcedotcom/lwc-dev-mobile-core/pull/83#discussion_r1599216704).

PR #84, opened 2024-05-22
Add util function to generate web socket url for LDP

maliroteh-sf:


src/common/PreviewUtils.ts
```diff
@@ -128,4 +129,44 @@ export class PreviewUtils {
...
+
+        if (CommandLineUtils.platformFlagIsAndroid(platform)) {
+            return `wss://10.0.2.2:${port}`;
+        }
+
+        if (process.platform !== 'darwin') {
+            return `ws://localhost:${port}`; // cannot be Safari since it is only available on Mac
+        }
+
+        // Check to see if the default browser is Safari
```
khawkins, line 164: Nit: Can you preface this comment with something like "macOS use case:"? Just to make it clear that, in this block, we're working in a macOS environment.

src/common/PreviewUtils.ts
```diff
@@ -128,4 +129,44 @@ export class PreviewUtils {
...
+        // From https://apple.stackexchange.com/questions/313454/applescript-find-the-users-set-default-browser
+        const cmd =
+            "defaults read ~/Library/Preferences/com.apple.LaunchServices/com.apple.launchservices.secure | awk -F'\"' '/http;/{print window[(NR)-1]}{window[NR]=$2}'";
```
khawkins, line 168: The thing I'm not wild about with this approach is making `awk` do the post-processing—mostly because of how it's not easy to read, maintain, or test (as `awk` tends to be).

I wonder if we shouldn't use something like [`plutil`](https://stackoverflow.com/a/63790334) to put what otherwise comes out of the `defaults read` output into a parseable format, and parse it ourselves? May be overkill though.

khawkins (commented):
Just the one comment on parsing the macOS defaults, but this approach seems fine to me overall.

src/common/PreviewUtils.ts
maliroteh-sf: As you said, it might be overkill. I thought about using packages like [default-browser-id](https://www.npmjs.com/package/default-browser-id) or [default-browser](https://www.npmjs.com/package/default-browser) but thought that those too might be over kill?

We only really care about whether on macOS the default browser is Safari or not. Looking at [its code](https://www.npmjs.com/package/default-browser-id?activeTab=code), `default-browser-id` is the most light-weight one but still, is it worth pulling this one in vs just use `awk` ?

We could always update the logic here to use something else in the future if need be?

khawkins: > We could always update the logic here to use something else in the future if need be?

Let's go with that. It's a minor concern, and mostly due to the fact that it's not the greatest insight into the plist data that's returned. But there's not much else to be gained from that plist data for our use case, anyway, so... ¯\\_(ツ)_/¯ yeah let's punt.

khawkins (approved):

sfdctaka (approved):
LGTM!

PR #87, opened 2024-05-30
Fix some unit tests

maliroteh-sf:


test/unit/common/CommonUtils.test.ts
```diff
@@ -6,20 +6,73 @@
...
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    public _write(chunk: Buffer, encoding: string, callback: (error?: Error | null) => void): void {
+        data.push(chunk);
+        callback();
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    public close(callback?: ((err?: NodeJS.ErrnoException | null | undefined) => void) | undefined): void {
```
sfdctaka, line 37: Do the overrides need to have their parent's class's method?

maliroteh-sf: Good point, no they don't.

test/unit/common/CommonUtils.test.ts
```diff
@@ -300,4 +362,26 @@ describe('CommonUtils', () => {
             parentPath: fname
         };
     }
+
+    async function verifyDownloadFails(url: string, mockResponse: any) {
+        if (url.startsWith('https:')) {
+            httpsGetStub.callsFake((_, callback) => {
+                setTimeout(() => callback(mockResponse), 100);
+                return { on: () => {}, end: () => {} };
+            });
```
sfdctaka, line 370: nit: this callback can be refactored outside of `if` since it's also used below in `httpsGetStub.callsFake(...)`.

sfdctaka (approved):
Pulled the code and ran unit test on my machine. LGTM!

khawkins (approved):

PR #88, opened 2024-05-30
feat: add CryptoUtils to generate self-signed cert

maliroteh-sf:


.yarnrc.txt
```diff
@@ -0,0 +1 @@
+network-timeout 600000
```
maliroteh-sf, line 1: In CI on Windows machines, sometimes `yarn install` keeps timing out and restarting. In `lwc-dev-mobile` we used this `.yarnrc` file to work around that so adding it here too.

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,150 @@
...
+                        value: '127.0.0.1'
+                    },
+                    {
+                        type: 7, // IP
+                        value: '10.0.2.2'
+                    },
+                    {
+                        type: 7, // IP (IPv6 loopback)
+                        value: '::1'
+                    }
```
maliroteh-sf, line 138: regardless of whether `localhost` is being used for `hostname` or not, I think we should always white-list these special IPs..... any objections?

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
+
+const messages = Messages.loadMessages('@salesforce/lwc-dev-mobile-core', 'crypto-utils');
+
+export type PEMCertificate = {
+    derCertificate: string;
+    pemCertificate: string;
+    pemPrivateKey: string;
+    pemPublicKey: string;
+};
```
maliroteh-sf, line 19: NOTE: iOS and Android need cert in DER format

I've been debating whether only have a generic `certificate: string` field here rather than both `pem` and `der` certificates.

If we choose to have a generic field then we need to to add another param to `generateSelfSignedCert` to indicate whether we want PEM or DER format + add extra code logic around using the param to generate the rightly formatted cert.

But in my experience so far in using `node-forge` it seems to be really fast. Bulk of execution time is taken by `createCertificate()` and `sign()` methods of `node-forge` (which still is under 1 sec).... once the signed cert is created, converting it to PEM or DER seems to be super fast.

Because of the fast speed of `node-forge` I oped to generate both PEM and DER formats of the cert at the same time and let the caller consume whichever it needs.

If we don't like this then I can update to only have a generic `certificate: string` field here and pass an extra param to `generateSelfSignedCert`

khawkins: I have a few thoughts on this:

- I think it's fine to represent both formats in the same data structure.
- I think we should call the data structure itself something different than `PEMCertificate` in that case, like `SSLCertificateData` or somesuch.
- I don't know about Android, but I am able to drag and drop PEM certificates into iOS simulators and have that "just work".
- `derCertificate` should likely not be represented as a `string`, as it's a binary data structure to represent the certificate.

maliroteh-sf: I just finished testing on Android (API 24 - 34) and it works with both PEM and DER.

I was unable to use PEM on iOS via drag/drop the PEM file generated by the code in here. My simulator gives the error `Simulator device failed to complete the requested operation`

But instead of drag/drop if I open safari on mobile and then navigate to `file://path/to/cert` then it installs the cert just fine. And since I was unable to find a way to auto-install the cert on iOS, we'll have to go the manual route which means using mobile safari so I'll go ahead and remove the DER support from this PR.

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+        } else {
+            endDate.setDate(startDate.getDate() + validity);
+        }
+
+        cert.publicKey = keys.publicKey;
+        cert.serialNumber = '01';
+        cert.validity.notBefore = startDate;
+        cert.validity.notAfter = endDate;
+
+        const attrs = [
```
khawkins, line 70: I'm reluctant to set values here that we don't need, and definitely want to be careful about how we represent the certificate configuration, so as not to overstate its "authoritative" value. This is a self-signed cert that shouldn't be used outside of development scenarios, at the end of the day. Here are my thoughts on changes here:

- Leave out `shortName`, `localityName`, and `OU`. These values shouldn't be required to create a functioning certificate.
- `countryName` hard-coded to 'US' is probably fine for now.
- `organizationName` should be "Salesforce Local Dev Preview - Local Server Self-Signed Cert"

src/common/CryptoUtils.ts
khawkins: Ah I see. My functioning drag-and-drop PEM for iOS was for a certificate created with `openssl`, so there are probably some differences between that and this process. I'm good with whichever approach we take that makes it less clicks or typing for the user.

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+            { name: 'localityName', value: 'San Francisco' },
+            { name: 'organizationName', value: 'Salesforce Inc.' },
+            { shortName: 'OU', value: 'LocalDevPreview' }
+        ];
+
+        cert.setSubject(attrs);
+        cert.setIssuer(attrs);
+
+        // Add the subjectAltName and ExtendedKeyUsage extensions, which are required by Apple
+        cert.setExtensions([
```
khawkins, line 83: I'd like us to try to clean up the extensions data that we need and don't need. I'll make notes on each section.

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+            {
+                name: 'basicConstraints',
```
khawkins, line 85: This isn't a CA certificate. We should either omit this section entirely, or set `CA` to `false`.

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+                cA: true
+            },
+            {
+                name: 'keyUsage',
```
khawkins, line 89: The only values I've ever needed to set in this section are `digitalSignature` and `keyEncipherment`.

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+                keyCertSign: true,
+                digitalSignature: true,
+                nonRepudiation: true,
+                keyEncipherment: true,
+                dataEncipherment: true
+            },
+            {
+                name: 'extKeyUsage',
```
khawkins, line 97: I don't think this section is necessary.

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+                serverAuth: true,
+                clientAuth: true,
+                codeSigning: true,
+                emailProtection: true,
+                timeStamping: true
+            },
+            {
+                name: 'nsCertType',
```
khawkins, line 105: These configurations are deprecated and not necessary.

src/common/CryptoUtils.ts
khawkins: Oh maybe for macOS? I know I get similar prompts when I import keys into the Keychain on my computer.

src/common/CryptoUtils.ts
khawkins: I think this is a good idea, and basically kills two birds with one stone with respect to Android emulators. Works for me.

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,155 @@
...
+                        value: '10.0.2.2'
+                    },
+                    {
+                        type: 7, // IP (IPv6 loopback)
+                        value: '::1'
+                    }
+                ]
+            },
+            {
+                name: 'subjectKeyIdentifier'
```
khawkins, line 136: Weird that this doesn't have a value. From https://www.openssl.org/docs/man1.0.2/man5/x509v3_config.html#Subject-Key-Identifier I would have expected a value of `hash`. ¯\\_(ツ)_/¯

src/common/CryptoUtils.ts
maliroteh-sf: I put most of these here b/c I saw it in the examples provided by `node-forge` and as you pointed out we probably can remove most of these altogether.

For the cert to be valid on iOS it needs to meet [these requirements](https://support.apple.com/en-gb/103769).

src/common/CryptoUtils.ts
maliroteh-sf: > derCertificate should likely not be represented as a string, as it's a binary data structure to represent the certificate.

It actually is. We generate the DER cert using this line:
```
const derCert = forge.asn1.toDer(forge.pki.certificateToAsn1(cert)).getBytes();
```

The call to `getBytes()` actually returns a string but when we are saving the content to file we should just use `fs.writeFileSync(path, derCertificate, { encoding: 'binary' })`

src/common/CryptoUtils.ts
maliroteh-sf: I removed it

src/common/CryptoUtils.ts
khawkins: Oh sorry, I didn't follow the definition through—just going on what I knew about the format.

src/common/CryptoUtils.ts
maliroteh-sf: I removed it

src/common/CryptoUtils.ts
maliroteh-sf: I removed other values

src/common/CryptoUtils.ts
khawkins: > For the cert to be valid on iOS it needs to meet [these requirements](https://support.apple.com/en-gb/103769).

Ah, didn't know about the newer EKU requirements. They don't seem to impact my usage, without them. But I'm fine with going with those guidelines, since they're out there and all. We could probably just get away with the `serverAuth` entry, out of the ones we have at the moment, per their requirements.

src/common/CryptoUtils.ts
maliroteh-sf: If I remove this section entirely or if I set `CA = false` then after installing the cert on my sim and going to `Settings > General > VPN & Device Management > localhost` it shows that the certificate is not verified. But by including this section and having `CA = true`, it would show as verified.

So you're positive that we don't need it?
Screenshot 2024-05-31 at 4 36 00 PM

khawkins: My experience is a bit apples and oranges, as I'm typically creating these certificates with `openssl`. We can leave it for now, and tweak later as necessary.

src/common/CryptoUtils.ts
maliroteh-sf: Updated as you requested.

src/common/CryptoUtils.ts
```diff
@@ -0,0 +1,132 @@
...
+
+        cert.publicKey = keys.publicKey;
+        cert.serialNumber = '01';
+        cert.validity.notBefore = startDate;
+        cert.validity.notAfter = endDate;
+
+        const attrs = [
+            { name: 'commonName', value: hostname },
+            { name: 'countryName', value: 'US' },
+            { name: 'organizationName', value: 'Salesforce Inc.' }
```
khawkins, line 73: I want `organizationName` to be something that reflects that this is just a development-environment certificate. So something like "Salesforce Local Development Server — Self-Signed Cert".

khawkins (approved):
Other than changing the `organizationName` value to be more dev-centric, this looks good to me.

khawkins (approved):

PR #89, opened 2024-06-03
refactor: define platform as enum

maliroteh-sf:


src/common/Common.ts
```diff
@@ -57,26 +57,28 @@ export class SetUtils {
...
 
+export enum Platform {
+    desktop = 'desktop',
+    ios = 'ios',
+    android = 'android'
+}
+
 export enum FlagsConfigType {
-    Platform,
+    PlatformType,
```
maliroteh-sf, line 67: Had to rename this one to avoid clashing with `Platform` enum

src/common/Common.ts
```diff
@@ -142,18 +129,16 @@ export class CommandLineUtils {
...
+                    platform: Flags.option({
                         char: 'p',
-                        description: supportsDesktop
-                            ? messages.getMessage('platformFlagIncludingDesktopDescription')
-                            : messages.getMessage('platformFlagMobileOnlyDescription'),
-                        required: isRequired,
-                        validate: supportsDesktop
-                            ? CommandLineUtils.validatePlatformFlagIncludingDesktop
-                            : CommandLineUtils.validatePlatformFlagMobileOnly
-                    })
```
maliroteh-sf, line 156: No need to define a custom message. For flags of type `option` the help message automatically adds the list of available options so makes our code simpler

src/common/Common.ts
khawkins: Can we rename this to `PlatformFlag` instead? My gut check was wondering what the difference between `PlatformType` and `Platform` was, which to me means the names are still too closely related. (I do understand the difference after checking it out. 🙂)

src/common/Common.ts
```diff
@@ -142,18 +129,16 @@ export class CommandLineUtils {
...
+                        description: messages.getMessage('platformFlagDescription'),
```
sfdctaka, line 136: Do you still not want to say it in here
image

Something like "Specify a platform: iOS, Android, or Desktop"? Or is it made generic so that more platforms are added in the future?

khawkins (approved):
Other than the minor renaming request, looks good to me. Nice that the CLI will handle the validation!

src/common/Common.ts
maliroteh-sf: It is made generic b/c `SfCommand` now automatically adds the list of options:
```
  -p, --platform=<option>  (required) Specify platform.
                           <options: ios|android>
```

PR #91, opened 2024-06-05
Export CryptoUtils + some cleanup

maliroteh-sf:


src/common/AndroidLauncher.ts
```diff
@@ -88,16 +88,24 @@ export class AndroidLauncher {
...
+
+                    if (address) {
+                        targetAppArguments.push({ name: PreviewUtils.SERVER_ADDRESS_PREFIX, value: address });
+                    }
+
+                    if (port) {
+                        targetAppArguments.push({ name: PreviewUtils.SERVER_PORT_PREFIX, value: port });
+                    }
+
+                    return AndroidUtils.launchAppInBootedEmulator(
```
maliroteh-sf, line 103: Cleanup the logic so that `launchApp` would only take care of launching the app with a given set of launch args.

Previously, `AndroidLauncher` & `IOSLauncher` contained some logic around determining values for a subset of launch arguments and `launchApp` contained some logic around determining values for another subset of launch arguments.

Now `AndroidLauncher` & `IOSLauncher` will produce `all` of the launch arguments and `launchApp` just consumes them

src/common/AndroidUtils.ts
```diff
@@ -47,16 +46,6 @@ export class AndroidUtils {
...
-     * Converts a path to UNIX style path.
-     *
-     * @param dirPath Input path.
-     * @returns UNIX style path.
-     */
-    public static convertToUnixPath(dirPath: string): string {
-        return dirPath.replace(/[\\]+/g, '/');
-    }
-
     /**
```
maliroteh-sf, line 49: This is not Android specific and should have been in `CommonUtils` instead so moved it there.

src/common/CryptoUtils.ts
```diff
@@ -70,7 +70,7 @@ export class CryptoUtils {
         const attrs = [
             { name: 'commonName', value: hostname },
             { name: 'countryName', value: 'US' },
-            { name: 'organizationName', value: 'Salesforce Local Development Server — Self-Signed Cert' }
+            { name: 'organizationName', value: 'Salesforce Local Development Server Self Signed Cert' }
```
maliroteh-sf, line 73: Removing `-` b/c iOS seems to not like it

test/unit/common/AndroidLauncher.test.ts
```diff
@@ -32,7 +33,7 @@ describe('Android Launcher tests', () => {
         const launcher = new AndroidLauncher('Pixel XL');
         await launcher.launchPreview('helloWorld', '~', undefined, 'browser', undefined, '3333');
 
-        expect(launchUrlMock.calledWith('http://10.0.2.2:3333/lwc/preview/c/helloWorld', 5572));
+        expect(launchUrlMock.calledWith('http://10.0.2.2:3333/lwc/preview/c/helloWorld', 5572)).to.be.true;
```
maliroteh-sf, line 36: I didn't know that `expect(someBoolVar)` doesn't actually check whether `someBoolVar === true` Instead it just ensures that `someBoolVar` is not `undefined`. So updated `all` of out unit tests to explicitly do the check.

khawkins (approved):

PR #94, opened 2024-06-28
Update CryptoUtils to return buffer for DER cert + some refactoring

maliroteh-sf:


src/common/AndroidEnvironmentRequirements.ts
```diff
@@ -77,7 +77,7 @@ export class Java8AvailableRequirement implements Requirement {
      * Verifies that prerequisites for Android SDK are met and that Java 8 is installed.
      */
     public async checkFunction(): Promise<string> {
-        return AndroidUtils.androidSDKPrerequisitesCheck()
+        return AndroidUtils.androidSDKPrerequisitesCheck(this.logger)
```
maliroteh-sf, line 80: This is something that I've always wanted to do but kept postponing. Rather than having `static` loggers in multiple places, we should just update method signatures to take an optional logger. I think this is cleaner. Not to mention that, we had a couple of utility classes with static loggers, but these loggers were not being initialized in `LoggerSetup.ts` which was a bug. Getting rid of these static loggers will also help us avoid such issues in the future .

src/common/AndroidLauncher.ts
```diff
@@ -41,9 +41,10 @@ export class AndroidLauncher {
         targetApp: string,
         appConfig: AndroidAppPreviewConfig | undefined,
         serverPort: string,
-        targetingLwrServer: boolean = false
+        targetingLwrServer: boolean = false,
+        logger?: Logger
     ): Promise<void> {
-        const preferredPack = await AndroidUtils.fetchSupportedEmulatorImagePackage();
+        const preferredPack = await AndroidUtils.fetchSupportedEmulatorImagePackage(undefined, logger);
```
maliroteh-sf, line 47: Some of our methods currently take an optional `apiLevel`.... but we don't need `apiLevel` and have a separate work item for removing `apiLevel` and any associated logic from our code so later this will get clean up more.

src/common/AndroidLauncher.ts
```diff
@@ -53,18 +54,18 @@ export class AndroidLauncher {
...
-                    return AndroidUtils.createNewVirtualDevice(emuName, emuImage, androidApi, device, abi);
+                    return AndroidUtils.createNewVirtualDevice(emuName, emuImage, androidApi, device, abi, logger);
                 }
                 CommonUtils.updateCliAction(messages.getMessage('foundDeviceStatus', [emuName]));
                 return Promise.resolve();
             })
             .then(() => {
                 CommonUtils.updateCliAction(messages.getMessage('startDeviceStatus', [emuName]));
-                return AndroidUtils.startEmulator(emuName);
+                return AndroidUtils.startEmulator(emuName, false, true, logger);
```
maliroteh-sf, line 68: I with typescript would support calling functions using named parameter rather than me just passing `true` and `false`

src/common/AndroidEnvironmentRequirements.ts
khawkins: I don't know that I think it is cleaner. But I also think this is because our code base is largely functional, as opposed to OO with separations of concerns and responsibilities. In an OO world we'd create objects that hold onto common configuration such as this. But we're pretty far from that, so it's fair to embrace the functional nature of our project and take this approach. 👍

khawkins (approved):

PR #95, opened 2024-07-29
Generate web socket url using http and https ports

maliroteh-sf:


sfdctaka (approved):
LGTM!

src/common/AndroidLauncher.ts
```diff
@@ -75,9 +75,11 @@ export class AndroidLauncher {
                 if (PreviewUtils.isTargetingBrowser(targetApp)) {
                     let url = '';
                     if (targetingLwrServer) {
+                        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
khawkins, line 78: I'm taking these disablements to mean that `targetingLwrServer` is somehow implying that `address` and `port` will always be defined as strings in that use case? Can we express that in a different way than disabling linting warnings?

maliroteh-sf: This is the old code path (previewing using old dev server). Upgrading eslint dependencies now shows these as warnings but rather than spending time refactoring the code to properly get rid of the lint warnings I decided to just disable them since in the next release cycle we plan on removing all of the old code paths anyway.

khawkins (approved):

PR #96, opened 2024-07-30
Use ip instead of value

maliroteh-sf:


src/common/CommonUtils.ts
```diff
@@ -201,16 +201,12 @@ export class CommonUtils {
      * @param stdioOptions Options to be used for [stderr, stdin, stdout]. Defaults to ['ignore', 'pipe', 'ignore']
      * @returns Result of the command execution.
      */
-    public static executeCommandSync(
-        command: string,
-        stdioOptions: StdioOptions = ['ignore', 'pipe', 'ignore'],
-        logger?: Logger
-    ): string {
+    public static executeCommandSync(command: string, stdioOptions?: StdioOptions, logger?: Logger): string {
```
maliroteh-sf, line 204: A bug was introduced in https://github.com/forcedotcom/lwc-dev-mobile-core/pull/94 where it was possible to pass in `undefined` for `stdioOptions` which would then break the default behavior where it would use `['ignore', 'pipe', 'ignore']` when no stdio options are provided.

This minor change fixes the issue.

khawkins (approved):
🎉

PR #131, opened 2024-09-09
feat: refactor code for device discovery

maliroteh-sf:
### What does this PR do?
Refactors the codebase and moves device discovery logic from IOSUtils/AndroidUtils to classes created specifically for device operations such as discovery. This is the first PR in a series of work to be done to clean up the codebase and reduce the complexity of IOSUtils/AndroidUtils


src/common/AndroidLauncher.ts
```diff
@@ -89,6 +90,7 @@ export class AndroidLauncher {
                     CommonUtils.updateCliAction(messages.getMessage('launchAppStatus', [targetApp]));
 
                     const launchActivity = appConfig?.activity ?? '';
+                    const target = launchActivity ? `${targetApp}/${launchActivity}` : targetApp;
```
maliroteh-sf, line 93: Launching an app on a device is now simpler. It is done using a Device object, which under the hood will use iOS/Android utils. I made the code such that only a target is passed in. So on iOS the caller would need to pass in bundle id and on android the caller would pass in package name + target activity (rather than having a different method signature on Android to pass package name & target activity separately)

src/common/AndroidUtils.ts
```diff
@@ -558,13 +558,20 @@ export class AndroidUtils {
...
      */
     public static async rebootEmulator(portNumber: number, waitForBoot: boolean, logger?: Logger): Promise<void> {
-        return AndroidUtils.executeAdbCommand('shell reboot', portNumber, logger).then(() => {
-            if (waitForBoot) {
-                return AndroidUtils.waitUntilDeviceIsReady(portNumber, logger);
-            } else {
-                return Promise.resolve();
-            }
-        });
+        try {
```
maliroteh-sf, line 561: Our `reboot` command for Android has been broken for a while now (not sure exactly when it broke). `adb` sometimes returns an error even though it does reboot the device.... I changed the logic here to be more robust in rebooting.

src/common/Common.ts
```diff
@@ -18,6 +18,55 @@ import { CustomOptions, OptionFlag } from '@oclif/core/interfaces';
 Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
 const messages = Messages.loadMessages('@salesforce/lwc-dev-mobile-core', 'common');
 
+export class CaseInsensitiveStringMap {
```
maliroteh-sf, line 21: This is essentially a `Map` where you can index to it using `key` that is case insensitive... this is used when parsing adb output during device discovery phase and makes things more robust.

src/common/IOSUtils.ts
```diff
@@ -70,157 +75,6 @@ export class IOSUtils {
             .catch((error) => Promise.reject(new SfError(`The command '${command}' failed to execute ${error}`)));
     }
 
-    /**
```
maliroteh-sf, line 73: These are refactored and moved to `device/*` classes

src/common/device/AndroidDevice.ts
```diff
@@ -0,0 +1,120 @@
...
+export enum AndroidOSType {
+    googleAPIs = 'google_apis',
+    googlePlayStore = 'google_apis_playstore',
+    androidDesktop = 'android-desktop',
+    googleTV = 'google-tv',
+    androidWear = 'android-wear',
+    androidAutomotive = 'android-automotive'
+}
+
+export class AndroidDevice implements BaseDevice {
```
maliroteh-sf, line 22: First revision.... more to be added in follow-up PRs

src/common/device/AndroidDeviceManager.ts
```diff
@@ -0,0 +1,256 @@
...
+import path from 'node:path';
+import { Logger } from '@salesforce/core';
+import { AndroidUtils } from '../AndroidUtils.js';
+import { CaseInsensitiveStringMap, Version } from '../Common.js';
+import { CommonUtils } from '../CommonUtils.js';
+import { PlatformConfig } from '../PlatformConfig.js';
+import { AndroidDevice, AndroidOSType } from './AndroidDevice.js';
+import { DeviceType } from './BaseDevice.js';
+
+export class AndroidDeviceManager {
```
maliroteh-sf, line 18: First revision.... more to be added in follow-up PRs

src/common/device/AppleDevice.ts
```diff
@@ -0,0 +1,116 @@
...
+};
+
+export enum AppleOSType {
+    iOS = 'iOS',
+    tvOS = 'tvOS',
+    watchOS = 'watchOS',
+    xrOS = 'xrOS'
+}
+
+export class AppleDevice implements BaseDevice {
```
maliroteh-sf, line 31: First revision.... more to be added in follow-up PRs

src/common/device/AppleDeviceManager.ts
```diff
@@ -0,0 +1,196 @@
...
+ */
+
+import { Logger, SfError } from '@salesforce/core';
+import { Version } from '../Common.js';
+import { CommonUtils } from '../CommonUtils.js';
+import { PlatformConfig } from '../PlatformConfig.js';
+import { AppleDevice, AppleOSType, AppleRuntime } from './AppleDevice.js';
+import { DeviceType } from './BaseDevice.js';
+
+export class AppleDeviceManager {
```
maliroteh-sf, line 15: First revision.... more to be added in follow-up PRs

test/unit/common/AndroidLauncher.test.ts
```diff
@@ -51,8 +51,9 @@ describe('Android Launcher tests', () => {
 
         expect(
             launchAppMock.calledWith(
-                undefined,
+                5572,
```
maliroteh-sf, line 54: method signature changed so param orders changed too

test/unit/common/AndroidUtils.test.ts
```diff
@@ -364,7 +364,7 @@ describe('Android utils', () => {
             stdout: `${targetApp}/.MainActivity`
         });
 
-        await AndroidUtils.launchAppInBootedEmulator(undefined, targetApp, targetAppArgs, targetActivity, port);
+        await AndroidUtils.launchAppInBootedEmulator(port, `${targetApp}/${targetActivity}`, undefined, targetAppArgs);
```
maliroteh-sf, line 367: method signature changed so param orders changed too

test/unit/common/IOSUtils.test.ts
```diff
@@ -10,41 +10,16 @@ import { expect } from 'chai';
...
 import { IOSUtils } from '../../../src/common/IOSUtils.js';
 import { PreviewUtils } from '../../../src/common/PreviewUtils.js';
-import { IOSMockData } from './IOSMockData.js';
 
 describe('IOS utils tests', () => {
     const $$ = new TestContext();
     const DEVICE_TYPE_PREFIX = 'com.apple.CoreSimulator.SimDeviceType';
     const RUNTIME_TYPE_PREFIX = 'com.apple.CoreSimulator.SimRuntime';
 
-    const myCommandRouterBlock = (command: string): Promise<{ stdout: string; stderr: string }> => {
```
maliroteh-sf, line 20: Some tests were refactored and moved to other places

test/unit/common/IOSUtils.test.ts
```diff
@@ -173,7 +148,7 @@ describe('IOS utils tests', () => {
             ` ${PreviewUtils.PROJECT_DIR_ARG_PREFIX}=${projectDir}` +
             ' arg1=val1 arg2=val2 ';
 
-        await IOSUtils.launchAppInBootedSimulator(udid, undefined, targetApp, targetAppArgs);
+        await IOSUtils.launchAppInBootedSimulator(udid, targetApp, undefined, targetAppArgs);
```
maliroteh-sf, line 151: method signature changed so param orders changed too

src/common/device/AndroidDeviceManager.ts
```diff
@@ -0,0 +1,256 @@
...
+     *   { osType: AndroidOSType.googlePlayStore, minOSVersion: new Version(34, 0, 0) }, // Google APIs Play Store 34 or newer
+     *   { osType: AndroidOSType.androidWear, minOSVersion: new Version(3, 0, 0) }, // Android Wear 3 or newer
+     * ]
+     * ```
+     *
+     * You can also pass in `null` for osFilters in order to skip filtering and include all discovered devices.
+     *
+     * @returns An array of available devices filtered using the `osFilter` parameter.
+     */
+    public async enumerateDevices(
```
maliroteh-sf, line 46: With my new changes, our code can now handle devices of different types (phone, tablet, tv, auto, watch, vr, ...) But of course for now we only want to deal with phone/tablet so I've made the default values of the filters such that it would preserve the old behavior for backwards compatibility.

src/common/device/AppleDevice.ts
```diff
@@ -0,0 +1,116 @@
...
+        // Instead we need to shutdown and again start up a device.
+        await this.shutdown();
+        await this.boot(waitForBoot);
+    }
+
+    /**
+     * Attempts to shutdown the device.
+     */
+    public async shutdown(): Promise<void> {
+        await IOSUtils.shutdownDevice(this.id, this.logger);
```
khawkins, line 90: Do you think there's value to keeping these underlying methods in the utils classes? Or would it make more sense to just inline their implementations in this class? To me I guess it would depend on whether there's value to calling those methods _outside of_ this class.

khawkins (approved):
I think this is a great start to refactoring! I feel like it makes the code much more manageable. 🎉

src/common/device/AppleDevice.ts
maliroteh-sf: That's exactly what I was asking myself the other day :) I think I would like to move them out of the util classes in the near future. There are still other operations (like createDevice) that lives in the util classes and so when I get to moving that code in a follow up PR, I'll be moving these too (along with test coverage for these)

PR #133, opened 2024-09-10
feat: add ability to auto install a certificate on device

maliroteh-sf:
### What does this PR do?
Add ability to auto install a certificate on a device.

### DIFF
--- a/src/common/CryptoUtils.ts
+++ b/src/common/CryptoUtils.ts
@@ -139,7 +139,7 @@ export class CryptoUtils {
      * @returns A buffer containing the data of the certificate in DER format.
      */
     public static pemToDer(pemCertificate: string): Buffer {
-        const cert = this.getCertFromPEM(pemCertificate);
+        const cert = this.getCertFromPem(pemCertificate);
         const derCertString = forge.asn1.toDer(forge.pki.certificateToAsn1(cert)).getBytes();
         const derCertBuffer = Buffer.from(derCertString, 'binary');
         return derCertBuffer;
@@ -152,7 +152,7 @@ export class CryptoUtils {
      * @returns A string representing a certificate in PEM format.
      */
     public static derToPem(derCertificate: Buffer): string {
-        const cert = this.getCertFromDER(derCertificate);
+        const cert = this.getCertFromDer(derCertificate);
         const pemCert = forge.pki.certificateToPem(cert);
         return pemCert;
     }
@@ -166,8 +166,8 @@ export class CryptoUtils {
      */
     public static getSubjectHashOld(certData: SSLCertificateData): string {
         const cert = certData.derCertificate
-            ? this.getCertFromDER(certData.derCertificate)
-            : this.getCertFromPEM(certData.pemCertificate);
+            ? this.getCertFromDer(certData.derCertificate)
+            : this.getCertFromPem(certData.pemCertificate);
 
         // Get a reference to forge.pki.distinguishedNameToAsn1 by casting to ANY
         // We do this b/c @types/node-forge doesn't expose distinguishedNameToAsn1
@@ -205,8 +205,8 @@ export class CryptoUtils {
      */
     public static isExpired(certData: SSLCertificateData): boolean {
         const cert = certData.derCertificate
-            ? this.getCertFromDER(certData.derCertificate)
-            : this.getCertFromPEM(certData.pemCertificate);
+            ? this.getCertFromDer(certData.derCertificate)
+            : this.getCertFromPem(certData.pemCertificate);
         const now = new Date();
         return cert.validity.notAfter < now;
     }
@@ -221,14 +221,14 @@ export class CryptoUtils {
         return randomBytes(byteSize).toString('base64');
     }
 
-    private static getCertFromDER(derCertificate: Buffer): forge.pki.Certificate {
+    private static getCertFromDer(derCertificate: Buffer): forge.pki.Certificate {
         const derBytes = forge.util.decode64(derCertificate.toString('binary'));
         const asn1 = forge.asn1.fromDer(derBytes);
         const cert = forge.pki.certificateFromAsn1(asn1);
         return cert;
     }
 
-    private static getCertFromPEM(pemCertificate: string): forge.pki.Certificate {
+    private static getCertFromPem(pemCertificate: string): forge.pki.Certificate {
         const cert = forge.pki.certificateFromPem(pemCertificate);
         return cert;
     }
--- a/src/common/CryptoUtils.ts
+++ b/src/common/CryptoUtils.ts
@@ -6,7 +6,7 @@
  */
 import { randomBytes } from 'node:crypto';
 import forge from 'node-forge';
-import { Messages, SfError } from '@salesforce/core';
+import { Messages } from '@salesforce/core';
 
 Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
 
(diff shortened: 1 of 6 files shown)


src/common/CryptoUtils.ts
```diff
@@ -132,6 +132,72 @@ export class CryptoUtils {
...
+
+    /**
+     * Computes the subject hash (similar to subject_hash_old of OpenSSL). When invoking
+     * this method, the caller should either provide a DER or PEM certificate (but not both).
+     *
+     * @param derCertificate A buffer containing the data of the certificate in DER format.
+     * @param pemCertificate A string representing a certificate in PEM format.
+     * @returns A string representing the subject hash.
+     */
+    public static getSubjectHashOld(derCertificate?: Buffer, pemCertificate?: string): string {
```
maliroteh-sf, line 168: I could either have one function `getSubjectHashOld(derCertificate?: Buffer, pemCertificate?: string)` or break it up to two:
```
getDERSubjectHashOld(derCertificate: Buffer)
getPEMSubjectHashOld(pemCertificate: string)
```

Same for `isExpired()` function below. I don't have a particular preference for one over the other so will leave it up to the team to decide which approach to go with

src/common/CryptoUtils.ts
```diff
@@ -132,6 +132,72 @@ export class CryptoUtils {
...
+        if (!derCertificate && !pemCertificate) {
+            throw new SfError('Must provide either a DER or PEM certificate.');
+        } else if (derCertificate && pemCertificate) {
+            throw new SfError('Must provide either a DER or PEM certificate, but not both.');
+        }
+
+        const cert = derCertificate ? this.getCertFromDER(derCertificate) : this.getCertFromPEM(pemCertificate!);
```
sfdctaka, line 175: You could refactor lines 169-175 and 189-195 into one method?

maliroteh-sf: Sure but we first need to decide on the [above comment](https://github.com/forcedotcom/lwc-dev-mobile-core/pull/133/files#r1752904383), since that would dictate whether to even have these checks or not.

src/common/CryptoUtils.ts
```diff
@@ -132,6 +132,72 @@ export class CryptoUtils {
         };
     }
 
+    /**
+     * Converts a certificate from PEM to DER format.
+     *
+     * @param pemCertificate A string representing a certificate in PEM format.
+     * @returns A buffer containing the data of the certificate in DER format.
+     */
+    public static PEMtoDER(pemCertificate: string): Buffer {
```
khawkins, line 141: The method name is stretching the boundaries of proper lower camel case naming, which is the convention for JS/TS. I'd recommend `pemToDer()` or `pemToDER()`, but not the naming you have here.

src/common/CryptoUtils.ts
khawkins: I prefer having two method signatures. It will be much clearer to the consumer what the desired inputs should be, to get their desired outputs. I'd recommend that for anywhere that we're currently saying that "you need to provide this input parameter _or_ that input parameter, but not both."

khawkins: Another way to do it (I don't have a preference) would be to have a data structure that encapsulates _all_ formats that we support, e.g. `Certificate`, that has members like `pemRepresentation`, `derRepresentation`, etc. It opens the door to encapsulating conversion between them, into that umbrella class. But I'd prefer one of those two approaches, over the either/or method parameters.

src/common/CryptoUtils.ts
```diff
@@ -132,6 +132,72 @@ export class CryptoUtils {
...
+     * @returns A string representing a certificate in PEM format.
+     */
+    public static DERtoPEM(derCertificate: Buffer): string {
+        const cert = this.getCertFromDER(derCertificate);
+        const pemCert = forge.pki.certificateToPem(cert);
+        return pemCert;
+    }
+
+    /**
+     * Computes the subject hash (similar to subject_hash_old of OpenSSL). When invoking
```
khawkins, line 161: Nit:

```suggestion
     * Computes the old-style (MD5) subject hash (similar to subject_hash_old of OpenSSL) that Android uses to label its user certificates. When invoking
```

I went with some of the description convention that OpenSSL itself [uses](https://github.com/openssl/openssl/blob/e8498dc6455fc36f70dc3a0ca1ef82b34c088a90/apps/x509.c#L108-L111).

src/common/CryptoUtils.ts
```diff
@@ -141,4 +220,16 @@ export class CryptoUtils {
     public static generateIdentityToken(byteSize: number = 32): string {
         return randomBytes(byteSize).toString('base64');
     }
+
+    private static getCertFromDER(derCertificate: Buffer): forge.pki.Certificate {
```
khawkins, line 224: Nit: we should probably be consistent in naming convention with methods like [`derToPem()`](https://github.com/forcedotcom/lwc-dev-mobile-core/pull/133/files#diff-0712445dd2528d1fbb3eea9100be49d400ff2d2319caf17024b6f0414b4adba9R154), or vice versa. So either `getCertFromDer()` here, or `derToPEM()` there.

src/common/device/AndroidDevice.ts
```diff
@@ -103,18 +107,95 @@ export class AndroidDevice implements BaseDevice {
...
+     * Installs a certificate on the device.
+     *
+     * @param certData An SSLCertificateData object containing the certificate data.
+     */
+    public async installCert(certData: SSLCertificateData): Promise<void> {
+        // For Android emulators, we need to save the PEM certificate in a file named as the cert subject hash.
+        const subjectHash = CryptoUtils.getSubjectHashOld(certData);
+        const pemContent = certData.pemCertificate ?? CryptoUtils.derToPem(certData.derCertificate);
+        const fileName = `${subjectHash}.0`; // this is the special file name
+        const certFilePath = path.join(os.tmpdir(), fileName);
```
khawkins, line 185: Might be better to make use of `CommonUtils.createTempDirectory()` instead of `os.tmpdir()` here, as that's typically how we've standardized creating temporary files/folders in the project.

src/common/device/AndroidDevice.ts
```diff
@@ -103,18 +107,95 @@ export class AndroidDevice implements BaseDevice {
...
+        fs.writeFileSync(certFilePath, pemContent);
+
+        // We then need to push the file to the emulator (needs to be root-mountable).
+        await AndroidUtils.mountAsRootWritableSystem(this.id, this.logger); // boot with writable system access
```
khawkins, line 189: Will adding certs in this way require that users then boot their emulators with `-writable-system`, in order for the changes to persist?

src/common/device/AndroidDevice.ts
maliroteh-sf: With that approach we will have a unique directory created each time `CommonUtils.createTempDirectory()` is called. I would then need to write extra code logic around deleting that directory and the cert file in it, after cert is installed (or an error occurs).

But using `os.tmpdir()` directly means that the certfile (which is always named the same) will always be created in the same location (overwriting if it already exists) and so we don't need cleanup logic (the file is less than 1kb and is reused/overwritten in subsequent calls).

src/common/device/AndroidDevice.ts
maliroteh-sf: No It won't. After the cert is installed, if the user later on starts the device regularly (for example using Android Studio) the cert will still be there.

src/common/device/AndroidDevice.ts
```diff
@@ -103,18 +107,95 @@ export class AndroidDevice implements BaseDevice {
...
+            result = await AndroidUtils.executeAdbCommand(
+                `shell pm list packages | grep "${target}"`,
+                this.port,
+                this.logger
+            );
+        } catch {
+            // ignore and continue
+        }
+
+        return Promise.resolve(result ? true : false);
```
khawkins, line 128: TIL that `''` is falsy. I wasn't sure, but I checked it locally. Huh! ¯\\_(ツ)_/¯

khawkins (approved):

PR #136, opened 2024-09-19
update cert installation process

maliroteh-sf:
### What does this PR do?
Update cert installation process + make it more robust on Windows platform.


src/common/AndroidTypes.ts
```diff
@@ -147,148 +146,3 @@ export class AndroidPackage {
         }`;
     }
 }
-
-export class AndroidVirtualDevice {
```
maliroteh-sf, line 151: These are not needed anymore since they've been refactored into `AndroidDevice` and `AndroidDeviceManager`

src/common/AndroidUtils.ts
```diff
@@ -173,51 +172,6 @@ export class AndroidUtils {
         return Promise.resolve(PlatformConfig.androidConfig().supportedDeviceTypes);
     }
 
-    /**
-     * Attempts to fetch all available Android virtual devices.
-     *
-     * @returns An array of all available Android virtual devices.
-     */
-    public static async fetchEmulators(logger?: Logger): Promise<AndroidVirtualDevice[]> {
-        let devices: AndroidVirtualDevice[] = [];
```
maliroteh-sf, line 181: These are not needed anymore since they've been refactored into `AndroidDevice` and `AndroidDeviceManager`

src/common/device/AndroidDevice.ts
```diff
@@ -19,12 +20,12 @@ Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
 const messages = Messages.loadMessages('@salesforce/lwc-dev-mobile-core', 'common');
 
 export enum AndroidOSType {
-    googleAPIs = 'google_apis',
-    googlePlayStore = 'google_apis_playstore',
-    androidDesktop = 'android-desktop',
-    googleTV = 'google-tv',
-    androidWear = 'android-wear',
-    androidAutomotive = 'android-automotive'
+    googleAPIs = 'google apis',
```
maliroteh-sf, line 23: Leave it to Google to mess this up. When querying AVDs on different platforms, some of these come back with _ and some with - as work separators. To I updated the logic during parsing to just replace all _ and - with blank space and then we can do a consistent string check on different platforms and for different OS types

test/unit/common/AndroidMockData.ts
```diff
@@ -6,86 +6,113 @@
  */
 export class AndroidMockData {
     public static mockRawPackagesString = `Warning: File test string.........       
-    Installed packages:=====================] 100% Computing updates...
```
maliroteh-sf, line 9: updating mock data for more complete testing

src/common/AndroidUtils.ts
```diff
@@ -777,12 +501,12 @@ export class AndroidUtils {
      * @param numRetries Optional parameter that indicates the number of times the command will be executed again if ADB comes back with an error. Defaults to AndroidConfig.AdbNumRetryAttempts.
      * @param retryDelay Optional parameter that indicates the milliseconds of delay before retrying again. Defaults to AndroidConfig.AdbRetryAttemptDelay.
      */
-    public static async executeAdbCommandWithRetry(
+    public static async executeAdbCommand(
```
maliroteh-sf, line 504: I'm making a change to run ALL `adb` commands with retry mechanism. I don't know why we didn't do this in the first place along time ago. It's just a good thing to do since on Windows `adb` barf a lot and we need to run a command again for it to pass.

So by default an `adb` command now will be retried. If the caller really doesn't want the retry mechanism then they can pass in 0 for `numRetries`

src/common/device/AndroidDevice.ts
```diff
@@ -5,22 +5,33 @@
...
+    googlePlayStore = 'google apis playstore',
+    androidDesktop = 'android desktop',
+    googleTV = 'google tv',
+    androidWear = 'android wear',
+    androidAutomotive = 'android automotive'
+}
+
+export enum BootMode {
+    normal = 'normal',
+    systemWritablePreferred = 'systemWritablePreferred',
```
khawkins, line 33: Where would we end up using "preferred" vs. "mandatory", out of curiosity?

src/common/device/AndroidDevice.ts
```diff
@@ -199,5 +235,156 @@ export class AndroidDevice implements BaseDevice {
...
+     * @param bootMode Optional enum indicating the boot mode. Defaults to Normal.
+     * @param coldBoot Optional boolean indicating whether we should perform a cold boot. Defaults to false.
+     * @param waitForBoot Optional boolean indicating whether it should wait for the device to finish booting up. Defaults to true.
+     * @returns The ADB port that the emulator was launched on.
+     */
+    private async startEmulator(bootMode = BootMode.normal, coldBoot = false, waitForBoot = true): Promise<void> {
+        const port = await AndroidUtils.emulatorHasPort(this.id, this.logger);
+        const resolvedPortNumber = port ? port : await AndroidUtils.getNextAvailableAdbPort(this.logger);
+
+        if (resolvedPortNumber === port) {
```
khawkins, line 313: Nit: I would suggest pulling this boolean comparison out to a more meaningful name for this context. E.g.:

```typescript
const emulatorAlreadyBooted = (resolvedPortNumber === port);
if (emulatorAlreadyBooted) { // ...
```
Just to add clarity of how what this condition means to the casual glance of the maintainer.

src/common/device/AndroidDevice.ts
```diff
@@ -199,5 +235,156 @@ export class AndroidDevice implements BaseDevice {
...
+            msgKey = writable ? 'emulatorRelaunchWritableStatus' : 'emulatorRelaunchNotWritableStatus';
+        } else {
+            msgKey = writable ? 'emulatorLaunchWritableStatus' : 'emulatorLaunchNotWritableStatus';
+        }
+
+        CommonUtils.updateCliAction(messages.getMessage(msgKey, [this.id, resolvedPortNumber]));
+
+        // We intentionally use spawn and ignore stdio here b/c emulator command can
+        // spit out a bunch of output to stderr where they are not really errors. This
+        // is specially true on Windows platform. So instead we spawn the process to launch
```
khawkins, line 341: Nit:
```suggestion
        // is especially true on the Windows platform. So instead we spawn the process to launch
```

src/common/device/AndroidDevice.ts
```diff
@@ -130,14 +146,25 @@ export class AndroidDevice implements BaseDevice {
         return Promise.resolve(result?.trim() ? true : false);
     }
 
+    /**
+     * Attempts to install a native app on the device.
+     *
+     * @param appBundlePath Path to the app bundle of the native app.
+     */
+    public async installApp(appBundlePath: string): Promise<void> {
+        const pathQuote = process.platform === 'win32' ? '"' : "'";
```
khawkins, line 155: Nit: We also have this functionality [here](https://github.com/forcedotcom/lwc-dev-mobile-core/blob/2b97e570c2fc8352d81cfc76fcd19826113dc6eb/src/common/AndroidUtils.ts#L723). Perhaps we should define a public method for it in `AndroidUtils`, just to keep it consistent?

khawkins (approved):
Seems good, worth testing out! 👍

src/common/device/AndroidDevice.ts
maliroteh-sf: For local dev preview on mobile, the first step would be to boot the device, then see if it has cert installed or not, and if not then install the cert (but to install the cert we need the device to be booted with write system access).

So the LDP CLI command first attempts to boot a device with `preferred` writable-system. If the device is not already booted then we boot it with writable system right off the bat assuming that there may be a need to install the cert and in order to avoid a forced reboot into writable system later one, in order to save time. Even if the device already has the cert installed, booting it in writable system doesn't cause any issues with local dev previewing.

If the device is already booted (for example by AS) then we just continue and won't force-reboot into writable-system. Then we check to see if cert is installed or not. If it is installed then all is good and we continue to use the already booted device as is. But if it is not installed then we force-reboot into writable system.

So having this "preferred" vs. "mandatory" can help us get to local preview faster under certain conditions (b/c of less device reboot)

PR #151, opened 2024-12-19
ci: enable auto-merging of dependabot PRs

maliroteh-sf:


.github/workflows/auto-merge.yml
```diff
@@ -0,0 +1,61 @@
...
+
+                      // When dependabot creates PRs, it adds 'dependencies' as a label to the PRs.
+                      // Here we check to see if 'dependencies' label is present
+                      const hasDependenciesLabel = labels.some(label => label.name === 'dependencies');
+
+                      if (!hasDependenciesLabel) {
+                        core.setFailed("The 'dependencies' label is missing.");
+                      }
+
+            - name: Check if CI passed
```
ben-zhang-at-salesforce, line 29: my understanding that flow as  'on: pull_request: types: - labeled' would be triggered when pr is added a label.  I guess dependabot creates the pr with label in it,  so would 'Check if CI passed' always be unsuccessful? might having a time issue here.

maliroteh-sf: The way I understood is that `auto-merge` job will get cancelled if the label `dependencies` is not on the PR. If the label is there then we move to the next phase which is verifying the PR check statuses. If all of the checks have passed too, only then we attempt to merge.

maliroteh-sf: I don't think there will be timing issue because these are all defined as workflow steps and afaik the steps are executed in series not parallel.

.github/workflows/auto-merge.yml
```diff
@@ -0,0 +1,61 @@
...
+              uses: actions/github-script@v6
+              id: check-label
+              with:
+                  script: |
+                      const { context } = require('@actions/github');
+                      const { labels } = context.payload.pull_request;
+
+                      // When dependabot creates PRs, it adds 'dependencies' as a label to the PRs.
+                      // Here we check to see if 'dependencies' label is present
+                      const hasDependenciesLabel = labels.some(label => label.name === 'dependencies');
```
khawkins, line 23: The `dependencies` label gets added to both the roll-up minor/patch PRs _and_ the major version updates. Is there further gating to ensure that this only happens for minor/patch roll-up PRs?
image

khawkins: From [these docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#enabling-automerge-on-a-pull-request), it looks like we could get more granular based on metadata that comes with the PRs. It also looks like we could do it a bit more streamlined if we leverage the `gh` CLI as part of the process. Might be worth a look.

maliroteh-sf: @khawkins Oh good point.... I forgot to limit it to minor & patch only. ChatGPT is suggesting to simply look at the title of PR to determine what type of update it is. Since we've already enabled grouping of dependabot PRs for minor & patch, we could simply test to see whether the PR subject includes `minor-and-patch group` and if so then proceed with auto-merge. What do you think?

khawkins: I think checking for `minor-and-patch group` in the subject could be sufficient. It looks like there's a lot of interesting metadata that comes back from the https://github.com/dependabot/fetch-metadata GHA, so check out the README there and see if that changes your mind about your approach. But otherwise, I think checking the subject should be fine.

PR #157, opened 2025-02-07
chore: use c8 for code coverage

sfdctaka:
### What does this PR do?
Using `c8` to get code coverage. `nyc` isn't working as we inherited from sfdx cli project template. We may go back to `nyc` if we find a path to get it working.

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -1,7 +1,9 @@
 {
-  "check-coverage": true,
-  "lines": 75,
-  "statements": 75,
-  "functions": 75,
-  "branches": 75
+    "check-coverage": true,
+    "lines": 75,
+    "statements": 80,
+    "functions": 75,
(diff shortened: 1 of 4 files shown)


.nycrc
```diff
@@ -1,7 +1,9 @@
...
-  "statements": 75,
-  "functions": 75,
-  "branches": 75
+    "check-coverage": true,
+    "lines": 75,
+    "statements": 80,
+    "functions": 75,
+    "branches": 75,
+    "temp-directory": "coverage",
+    "reporter": ["text", "html"]
```
sfdctaka, line 8: `c8` can use `nyc`'s config which makes it easy to go back to `nyc` once we find a path to get that working.

maliroteh-sf (approved):

PR #163, opened 2025-02-18
chore: add baseline coverage

sfdctaka:
### What does this PR do?
It seems like this [PR](https://github.com/forcedotcom/lwc-dev-mobile-core/commit/1296ed9b8f90ea8b146b4e23deafda1c631ac791) was the last commit that ever uploaded code coverage successfully. Since then many PRs have not been able to upload the coverage to codecov and hence losing the base coverage along the way. 

This PR is to re-establish the baseline again. Also, adding `lcov` as the report that produces `lcov.info` which is a code coverage reporting format.

Updating codecov github action to V5 as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
         "postpack": "shx rm -f oclif.manifest.json oclif.lock",
         "prepack": "yarn clean-all && npm run sf-prepack",
         "sf-prepack": "sf-prepack",
-        "test": "c8 mocha 'test/unit/**/*.test.ts'",
+        "test": "c8 mocha 'test/unit/**/*.test.ts' && c8 report --reporter=lcov",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -129,7 +129,7 @@
         "postpack": "shx rm -f oclif.manifest.json oclif.lock",
         "prepack": "yarn clean-all && npm run sf-prepack",
         "sf-prepack": "sf-prepack",
-        "test": "c8 mocha 'test/unit/**/*.test.ts'",
+        "test": "c8 mocha 'test/unit/**/*.test.ts' && c8 report --reporter=lcov",
```
sfdctaka, line 132: Previously `c8` was just generating report locally. By adding a step to `report` using `lcov` as a reporter it creates `lcov.info` file which gets uploaded to codecov along with related json files.

maliroteh-sf (approved):

PR #171, opened 2025-06-25
fix: update yarn.lock to pick up latest oclif/table 0.4.9

ben-zhang-at-salesforce:
### What does this PR do?
refresh the yarn.lock to pick up the latest @oclif/table@0.4.9

### What issues does this PR fix or reference?
saleforcedx-vscode-mobile dependency update needs this fix.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1523,9 +1523,9 @@
     registry-auth-token "^5.0.3"
 
 "@oclif/table@^0.4.6":
-  version "0.4.6"
-  resolved "https://registry.yarnpkg.com/@oclif/table/-/table-0.4.6.tgz#4e07fc11de9b8a3187a5ea43edce71c42811ab7e"
-  integrity sha512-NXh72vHHYnDrWPmVfh4i7kDydz3CXm/tSAr17fWhmWfMM+8jGn5uo6FXtvB0cd9s4skvDqzoRcsRwOeR73zIKA==
+  version "0.4.9"
+  resolved "https://registry.yarnpkg.com/@oclif/table/-/table-0.4.9.tgz#bf1057e523d948aad8578d4bb721009589bed1b4"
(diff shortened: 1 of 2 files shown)


yarn.lock
ben-zhang-at-salesforce, line 1: removed the @oclif/table@^0.4.6" section, and yarn install to only refresh @oclif/table and its own dependencies.

ben-zhang-at-salesforce: solved

ben-zhang-at-salesforce:
Hi @khawkins,  is the .circleci/config.yml required? checked other peer node repos, and they don't have them too.

maliroteh-sf:
> Hi @khawkins, is the .circleci/config.yml required? checked other peer node repos, and they don't have them too.

@ben-zhang-at-salesforce I don't see a `.circleci/config.yml` in the repo and I don't have it locally either. Where is this coming from? We use github actions not circleci

ben-zhang-at-salesforce:
> > Hi @khawkins, is the .circleci/config.yml required? checked other peer node repos, and they don't have them too.
> 
> @ben-zhang-at-salesforce I don't see a `.circleci/config.yml` in the repo and I don't have it locally either. Where is this coming from? We use github actions not circleci

checked around and found the circleci git app in the settings. might some git org owner enabled circleci for our repo? 
image

maliroteh-sf (approved):
CircleCI has now been disabled but it won't reflect on this PR so you can ignore that check and merge.

khawkins:
> This is really annoying... why would they enable CircleCI on all repos in the org and not allow us to configure it? We could add a `.circleci/config.yml` to the repo to basically no-op but that is not a clean approach. Kevin already pinged them so let's see if they can disable it for our repos.

Did I miss an issue with the CircleCI enablement? I thought it was just an annoyance. Was something not working because that app was enabled?

PR #175, opened 2025-07-01
feat: add util func to get supported device types

maliroteh-sf:
Add utility function to get the list of supported iOS device types + update dependencies.

This is part of the effort to move common logic from `lwc-dev-mobile` to `lwc-dev-mobile-core`


sfdctaka (approved):
LGTM!

PR #183, opened 2025-09-24
fix: add apilevel flag for 'setup' command in iOS platform

haifeng-li-at-salesforce:
# Add API Level Support for iOS Setup Command

## Summary

Adds support for the `--apilevel` flag to the iOS platform setup command, allowing users to specify a minimum iOS version when checking for available simulator runtimes.

## Changes

-   **Setup Command**: Modified to pass the `apilevel` parameter to `IOSEnvironmentRequirements`
-   **iOS Requirements**: Enhanced `SupportedSimulatorRuntimeRequirement` to use custom API level filtering
-   **Runtime Filtering**: When `apilevel` is provided, only iOS runtimes matching or exceeding the specified version are validated

## Usage

```bash
# Setup with specific iOS version requirement
./bin/dev.js force:lightning:local:setup -p ios -l 16.4

# Setup with default iOS version (existing behavior)
./bin/dev.js force:lightning:local:setup -p ios
```

## Implementation Details

-   `IOSEnvironmentRequirements` constructor now accepts optional `apiLevel` parameter
-   `SupportedSimulatorRuntimeRequirement` creates custom filter when `apiLevel` is specified
-   Calls `AppleDeviceManager.enumerateRuntimes()` with appropriate filter array
-   Maintains backward compatibility when no API level is specified

## Testing

-   Added 10+ comprehensive test cases covering:
    -   Setup command with/without API level for both platforms
    -   Constructor behavior with/without API level parameter
    -   Runtime enumeration with custom filtering
    -   Error handling for invalid API levels
    -   Various version format validation

### DIFF
--- a/src/common/IOSEnvironmentRequirements.ts
+++ b/src/common/IOSEnvironmentRequirements.ts
@@ -6,6 +6,8 @@
  */
 import { Logger, Messages, SfError } from '@salesforce/core';
 import { CommonUtils } from './CommonUtils.js';
+import { Version } from './Common.js';
+import { AppleOSType } from './device/AppleDevice.js';
 import { AppleDeviceManager } from './device/AppleDeviceManager.js';
 import { PlatformConfig } from './PlatformConfig.js';
 import { Requirement, RequirementList } from './Requirements.js';
@@ -16,11 +18,11 @@ const messages = Messages.loadMessages('@salesforce/lwc-dev-mobile-core', 'requi
 export class IOSEnvironmentRequirements implements RequirementList {
     public requirements: Requirement[] = [];
     public enabled = true;
-    public constructor(logger: Logger) {
+    public constructor(logger: Logger, apiLevel?: string) {
         this.requirements = [
             new SupportedEnvironmentRequirement(logger),
             new XcodeInstalledRequirement(logger),
-            new SupportedSimulatorRuntimeRequirement(logger)
+            new SupportedSimulatorRuntimeRequirement(logger, apiLevel)
         ];
     }
 }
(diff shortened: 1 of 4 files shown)


src/common/IOSEnvironmentRequirements.ts
```diff
@@ -128,7 +131,9 @@ export class SupportedSimulatorRuntimeRequirement implements Requirement {
         const deviceManager = new AppleDeviceManager(this.logger);
 
         return deviceManager
-            .enumerateRuntimes()
+            .enumerateRuntimes(
+                this.apiLevel ? [{ osType: AppleOSType.iOS, minOSVersion: Version.from(this.apiLevel)! }] : undefined
```
maliroteh-sf, line 135: Technically this should be changed to `this.apiLevel ? ... : null` You should use `null` instead of `undefined`

Take a look at the method signature of `AppleDeviceManager.enumerateDevices()` and you'll see that for the `osFilters` param it should either be an array of filters or `null`

haifeng-li-at-salesforce: If `null` is used instead of `undefined`, the default `osFilters` value within `public async enumerateRuntimes(
        osFilters: Array<{ osType: string; minOSVersion: Version }> | null = [
            { osType: AppleOSType.iOS, minOSVersion: Version.from(PlatformConfig.iOSConfig().minSupportedRuntime)! }
        ]
    )` will be ignored since `null` is a value. Therefore, the default behavior would be changed.

maliroteh-sf: Oh I see.... what you want to achieve is to use the `apiLevel` if it is provided, otherwise use the default behavior? In that case, passing `undefined` is correct.

PR #184, opened 2025-09-24
feat: return device state for getting device list

ben-zhang-at-salesforce:
### Currently the emulator or simulator running state is not returned when runs 
`sf force lightning local device list -p android`
`sf force lightning local device list -p ios`

Make it hard to see if which are running and have to call start device again and again.  
Adding the status could make the `sf force lightning local device *` more ergonomic to use.


ben-zhang-at-salesforce:
no needed.

PR #190, opened 2025-11-05
chore: update node version to >=22

ben-zhang-at-salesforce:


yarn.lock
```diff
@@ -4,23 +4,15 @@
 
 "@alcalzone/ansi-tokenize@^0.1.3":
   version "0.1.3"
-  resolved "https://registry.npmjs.org/@alcalzone/ansi-tokenize/-/ansi-tokenize-0.1.3.tgz"
+  resolved "https://registry.yarnpkg.com/@alcalzone/ansi-tokenize/-/ansi-tokenize-0.1.3.tgz#9f89839561325a8e9a0c32360b8d17e48489993f"
```
maliroteh-sf, line 7: We have `.npmrc` and `.yarnrc` in this repo that are pointing to `https://registry.npmjs.org`

If we're switching over to `https://registry.yarnpkg.com` then I guess we no longer need `.npmrc` and `.yarnrc` OR need to rename them to `.yarnrc.yml`

ben-zhang-at-salesforce: ai research it a bit, we are using yarn 1.22.22 which is yarn classic and seems yarn classic writes yarnpkg.com url even we have .yarnrc which may not take effect.   let me dig a bit more into repos like @salesforce/sf-plugins-core and see what's best approach

ben-zhang-at-salesforce: seems other plugins like https://github.com/salesforcecli/plugin-apex or @salesforce/sf-plugins-core has the .npmrc or .yarnrc

It's removed now the lock file will all have yarnpkg.com/**/*.*

khawkins:
Project-level npmrc's and yarnrc's are only required if you need to specifically override some default behavior. Shouldn't be necessary in any of our public repos. It's best to go the the package manager's default idioms, unless there's a pressing reason to override them.

maliroteh-sf (approved):

PR #197, opened 2026-02-20
feat: add telemetry support via Product Feature Tracking (PFT)

ben-zhang-at-salesforce:
## Summary
- Add `TelemetryEmitter` abstraction and `SfCliTelemetryEmitter` implementation that wraps `@salesforce/core` Lifecycle telemetry
- Integrate telemetry emission into `BaseCommand.init()` to track CLI command executions
- Configure observability settings (`enableO11y`, `o11yUploadEndpoint`, `productFeatureId`) in package.json
- Export new `Telemetry` module from the package entry point
- Add `skipLibCheck` to tsconfig for compatibility with new dependencies

## Test plan
- [x] Unit tests added for `SfCliTelemetryEmitter` (verifies payload structure and data spreading)
- [x] Unit tests added for `BaseCommand` telemetry integration (verifies emission during init)
- [x] All 192 existing tests pass
- [x] Coverage thresholds met (88% statements, 76% branches, 78% functions, 88% lines)

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+registry "https://registry.yarnpkg.com"
--- a/messages/common.md
+++ b/messages/common.md
@@ -152,4 +152,4 @@ Error collecting network hardware ports: %s
 
 # error:cli:unsupportedVersion
 
-This command requires @salesforce/cli version %s or later. You are running version %s. Please update by running: npm install -g @salesforce/cli@latest
+This command requires @salesforce/cli version %s or later. You are running version %s. Please update to a newer version.
--- a/messages/common.md
+++ b/messages/common.md
@@ -149,3 +149,7 @@ Comparing 2 codename versions is not supported.
 # error:network:hardware:port
 
 Error collecting network hardware ports: %s
+
+# error:cli:unsupportedVersion
+
+This command requires @salesforce/cli version %s or later. You are running version %s. Please update by running: npm install -g @salesforce/cli@latest
--- a/src/cli/commands/force/lightning/local/setup.ts
+++ b/src/cli/commands/force/lightning/local/setup.ts
@@ -46,6 +46,11 @@ export class Setup extends BaseCommand {
         return RequirementProcessor.execute(this.commandRequirements, this.jsonEnabled());
     }
 
+    protected getTelemetryEventName(): string {
+        // eslint-disable-next-line no-underscore-dangle
+        return `${this._commandName} executed`;
+    }
+
     protected populateCommandRequirements(): void {
         const requirements: CommandRequirements = {};
 
--- a/src/cli/commands/force/lightning/local/setup.ts
+++ b/src/cli/commands/force/lightning/local/setup.ts
@@ -46,11 +46,6 @@ export class Setup extends BaseCommand {
         return RequirementProcessor.execute(this.commandRequirements, this.jsonEnabled());
     }
 
-    protected getTelemetryEventName(): string {
-        // eslint-disable-next-line no-underscore-dangle
-        return `${this._commandName}.executed`;
-    }
-
     protected populateCommandRequirements(): void {
         const requirements: CommandRequirements = {};
 
--- a/src/cli/commands/force/lightning/local/setup.ts
+++ b/src/cli/commands/force/lightning/local/setup.ts
@@ -48,7 +48,7 @@ export class Setup extends BaseCommand {
 
     protected getTelemetryEventName(): string {
         // eslint-disable-next-line no-underscore-dangle
-        return `${this._commandName} executed`;
+        return `${this._commandName}.executed`;
     }
 
     protected populateCommandRequirements(): void {
--- a/src/common/BaseCommand.ts
+++ b/src/common/BaseCommand.ts
@@ -122,7 +122,7 @@ export abstract class BaseCommand extends SfCommand<unknown> implements HasRequi
                 return this.populateCommandRequirements();
             })
             .then(() => {
-                this.cmdTelemetryEmitter.emitTelemetry('lwc-dev-mobile-core', {
+                this.cmdTelemetryEmitter.emitTelemetry(this.getTelemetryEventName(), {
                     commandName: this.commandName
                 });
             });
@@ -233,4 +233,6 @@ export abstract class BaseCommand extends SfCommand<unknown> implements HasRequi
                 return LoggerLevel.WARN;
         }
     }
+
+    protected abstract getTelemetryEventName(): string;
 }
--- a/src/common/BaseCommand.ts
+++ b/src/common/BaseCommand.ts
@@ -5,14 +5,21 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/MIT
  */
 import { SfCommand } from '@salesforce/sf-plugins-core';
-import { Logger, LoggerLevel } from '@salesforce/core';
+import { Logger, LoggerLevel, Messages } from '@salesforce/core';
 import { z, toJSONSchema } from 'zod/v4';
+import semver from 'semver';
 import { CommandLineUtils } from './CommandLineUtils.js';
 import { HasRequirements, CommandRequirements } from './Requirements.js';
 import { OutputFormat } from './Common.js';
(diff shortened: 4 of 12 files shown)


messages/common.md
```diff
@@ -149,3 +149,7 @@ Comparing 2 codename versions is not supported.
 # error:network:hardware:port
 
 Error collecting network hardware ports: %s
+
+# error:cli:unsupportedVersion
+
+This command requires @salesforce/cli version %s or later. You are running version %s. Please update by running: npm install -g @salesforce/cli@latest
```
ben-zhang-at-salesforce: error ms if @salesforce/cli version < 2.123.1

src/common/BaseCommand.ts
```diff
@@ -74,7 +104,10 @@ export abstract class BaseCommand extends SfCommand<unknown> implements HasRequi
 
         return super
             .init()
-            .then(() => this.parse())
+            .then(() => {
+                BaseCommand.verifyCliVersion(this.config.version);
```
ben-zhang-at-salesforce, line 107: check 2.123.1 or above installed.

src/common/BaseCommand.ts
```diff
@@ -87,6 +120,11 @@ export abstract class BaseCommand extends SfCommand<unknown> implements HasRequi
                 logger.setLevel(logLevel);
                 this.cmdLogger = logger;
                 return this.populateCommandRequirements();
+            })
+            .then(() => {
+                this.cmdTelemetryEmitter.emitTelemetry('lwc-dev-mobile-core', {
```
ben-zhang-at-salesforce: this will trigger the pft event sending in [here](https://github.com/salesforcecli/plugin-telemetry/blob/785c34022c030d8dbadd32a479e5ed72ad885b0d/src/uploader.ts#L115)

the event look like below 

Image

.yarnrc
```diff
@@ -0,0 +1 @@
+registry "https://registry.yarnpkg.com"
```
ben-zhang-at-salesforce, line 1: add this to make sure `yarn install` will generate yarn.lock whose package url will be like https://registry.yarnpkg.com/... instead of npm.com .... if dev machine has different configuration.

messages/common.md
maliroteh-sf: Rather than including the command for updating, it might be best to just keep it generic like "Please update to a newer version."

I believe there are different ways of installing the salesforce CLI (via npm or via pkg installer) and even on different platforms (Windows, Linux, Mac, ...) the command format might be different.

src/common/BaseCommand.ts
maliroteh-sf: Using `'lwc-dev-mobile-core'` for eventName is incorrect. It should be camel-cased and not hyphenated. It should also be 2 parts and the second part is a verb (just like what we did for Nimbus... see discussion in slack thread).

Also note that when we define other commands (in other libraries) that extend from `BaseCommand`, when you execute any of those commands, you are still using `'lwc-dev-mobile-core'` as eventName which is wrong.

For example if you run the Device Create command from `lwc-dev-mobile` then we should not log that as `lwc-dev-mobile-core`.

For that reason perhaps we should use `this.commandName` as the eventName? But then again it is not guaranteed that `this.commandName` is in the correct syntax (camel cased and not hyphenated).

Perhaps something to discuss with the team and decide on, before considering to merge this PR.

ben-zhang-at-salesforce: the telemetry eventName here is not ignored by down stream. Downstream will send a proper PDP eventName as "salesforceCli.executed" as shown in the superset. 
image

the first method for emitTelemetry method need a eventName,  my latest commit change id to `${commandName} executed`

ben-zhang-at-salesforce: the event is sent from [this line ](https://github.com/salesforcecli/plugin-telemetry/blob/68437593561cb86bbbd26f2e93937d28e47d2665/src/uploader.ts#L163) from plugin-telemetry which is included by @salesforce/cli(version 2.123.1).

src/cli/commands/force/lightning/local/setup.ts
```diff
@@ -46,6 +46,11 @@ export class Setup extends BaseCommand {
         return RequirementProcessor.execute(this.commandRequirements, this.jsonEnabled());
     }
 
+    protected getTelemetryEventName(): string {
+        // eslint-disable-next-line no-underscore-dangle
+        return `${this._commandName} executed`;
```
maliroteh-sf: Based on what I read [here](https://docs.google.com/document/d/1HKx9lFkR8tv5IlGOexjvEONEXDzmiNb3gDJ0-mD0QaM/edit?tab=t.0#heading=h.lzz0ovi1q4df) and the case study from [here](https://docs.google.com/document/d/1638qMF8LeOzULg1YInUoks3KQwTAMYgtiisLGf3TtzY/edit?tab=t.0) it seems that they don't want to have spaces in the event name string. The string should be 2 part where the first part describes the feature/task and the second part describes the type of action (and action should be 1 word which is all lowercase). And the 2 parts are separated by a `.` (eg `part1.part2`)

Our command names are in the form of `force:lightning:local:setup` (i.e delimiter is `:` or blank space)

I think we should update this function to something like this:
```
protected getTelemetryEventName(): string {
    // remove : or spaces from the command name then camel-case the words
    let sanitizedCommandName = toCamelCase(this._commandName)
    return `${sanitizedCommandName}.executed`
}

private toCamelCase(input: string): string {
  // Split the input by colon or space
  const words = input.split(/[:\s]+/);

  if (words.length === 0) return '';

  // Keep first word lowercase, capitalize the rest
  const camelCased = words
    .map((word, index) => {
      const lower = word.toLowerCase();
      if (index === 0) return lower;
      return lower.charAt(0).toUpperCase() + lower.slice(1);
    })
    .join('');

  return camelCased;
}
```

src/common/Telemetry.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Lifecycle } from '@salesforce/core';
+
+export type TelemetryEmitter = {
+    emitTelemetry(eventName: string, data: Record<string, unknown>): void;
+};
+
+export class SfCliTelemetryEmitter implements TelemetryEmitter {
+    // eslint-disable-next-line class-methods-use-this
+    public emitTelemetry(eventName: string, data: Record<string, unknown>): void {
+        void Lifecycle.getInstance().emitTelemetry({
```
maliroteh-sf, line 16: When I build the changes in this PR and then run the `Setup - iOS` command, I see that the execution comes to this line but when I step into `Lifecycle.getInstance().emitTelemetry()` I see that the execution goes to `emit()` in `lifecycleEvents.js` and it errors out b/c the `telemetry` event is not registered. So as it is right now, the changes in this PR will not result in PFT log entries. Have you tested out the changes in this PR to see if entries are indeed logged?

Image

ben-zhang-at-salesforce: I verified `sf force lightning local setup -p android`  and found the event in the superset console.  let me try ios

ben-zhang-at-salesforce: debugged a bit, the o11yReport has caching enabled by default 
image

When the event flushed, I do see it in superset.

src/cli/commands/force/lightning/local/setup.ts
maliroteh-sf: Discussed with Ben offline.... no matter what we pass in as the eventName, it will be overwritten by the logger and only `saleforceCli.executed` will be logged.

maliroteh-sf: @ben-zhang-at-salesforce since the eventName is always overwritten, maybe we should make a slight change. In `BaseCommand` instead of having `protected abstract getTelemetryEventName(): string;` as an abstract method, we should just define it like this:
```
protected getTelemetryEventName(): string {
        // eslint-disable-next-line no-underscore-dangle
        return `${this._commandName}.executed`;
}
```

And then remove `getTelemetryEventName()` from `setup.ts`

This way, as we publish a new version of `lwc-dev-mobile-core` and consume it in other places, we don't need to repeatedly add a `getTelemetryEventName()` to each of the other commands since they will all inherit it from BaseCommand.

maliroteh-sf (approved):

PR #220, opened 2026-08-06
fix: harden shell-executing primitives against OS command injection (@W-23665289@)

ben-zhang-at-salesforce:
## What & Why

**W-23665289** — OS command injection via shell-executing primitives (`spawn({shell:true})` / string-based command runners).

Untrusted inputs (device names, device types, URLs, network interface names) were concatenated into shell command strings and executed through a shell, creating an OS command-injection surface. This PR structurally removes that class of bug by executing child processes with `shell:false` and passing untrusted values as inert `argv` elements. The few sinks that genuinely require a shell (pipes to `awk`/`grep`) now shell-quote their untrusted arguments as defense-in-depth.

## Changes

- **`CommonUtils.spawnWrapper`** — drop `shell:true`; the OS no longer performs shell parsing of arguments.
- **`CommonUtils.shellQuote`** (new) — POSIX single-quote / Windows double-quote wrapping for the sinks that must use a shell.
- **`CommonUtils.launchUrlInDesktopBrowser`** — build an `argv` array instead of a shell string.
- **`AndroidUtils.createNewVirtualDevice` / `spawnChild`** — build `argv` arrays, `shell:false`.
- **`AndroidDevice`** emulator launch — `argv` + `shell:false`, plus an `error` event handler (missing-binary robustness the change surfaced).
- **`IOSUtils` / `AppleDevice`** — route all `xcrun` sinks through `spawnCommandAsync` with `argv` arrays; shell-quote the two pipe sinks.
- **`MacNetworkUtils`** — shell-quote the `ifconfig | awk` device name.

## Testing

- `yarn build` ✓, `yarn lint` ✓ (2 pre-existing unrelated warnings in `AppleDeviceManager.ts`), **208 tests passing**.
- Added tests: `spawnWrapper` proves no shell interpretation (real subprocess, verbatim argv), `shellQuote` POSIX/Windows behavior, malicious-argument passthrough, and injection cases across Android/iOS sinks.
- Coverage above all thresholds (stmts 89.3%, branch 76.9%, funcs 80.6%, lines 89.3%).

The `--json` validation-bypass half of this ticket is fixed in a companion PR against `forcedotcom/lwc-dev-mobile`.

### DIFF
--- a/src/common/AndroidUtils.ts
+++ b/src/common/AndroidUtils.ts
@@ -42,6 +42,15 @@ export class AndroidUtils {
     private static sdkManagerCommand: string | undefined;
     private static sdkRoot: AndroidSDKRoot | undefined;
 
+    // Characters safe to pass to cmd.exe as (Node-quoted) argv elements. Allowlist rather than
+    // denylist: the values that reach the Windows cmd.exe path are structured (an already
+    // space-normalized emulator name, a device id, a `system-images;api;image;abi` path, an abi),
+    // so a strict allowlist rejects anything unexpected instead of trying to enumerate every cmd
+    // metacharacter. `;` and `=` are permitted because the system-image path uses `;` separators
+    // and name=value args exist elsewhere; space is permitted because callers other than AVD-name
+    // normalization may include it. Every cmd metacharacter (`& | < > ^ ( ) " % !`) is excluded.
+    private static readonly CMD_SAFE_ARG = /^[A-Za-z0-9 ._:;@\\/=+-]+$/;
+
     /**
      * Indicates whether Android packages are cached or not.
      *
@@ -876,17 +885,27 @@ export class AndroidUtils {
     // `.exe` but not `.bat`/`.cmd` (ENOENT), and post-CVE-2024-27980 Node refuses to spawn a
     // `.bat`/`.cmd` unless `shell:true` (EINVAL). We therefore launch it through `cmd.exe /c` with
     // shell:false — the same pattern launchUrlInDesktopBrowser uses for `start`. cmd.exe resolves
-    // `avdmanager` to `avdmanager.bat` via PATHEXT, and because spawn still runs with shell:false
-    // Node quotes each argv element; cmd.exe honors those double quotes for the RCE metacharacters
-    // (`& | ; < >`), so untrusted values are not re-interpreted as commands. We do NOT use
-    // shell:true here: shell:true does not escape argv (it joins with spaces), which would
-    // reintroduce the injection this PR removes.
+    // `avdmanager` to `avdmanager.bat` via PATHEXT.
+    //
+    // shell:false alone is NOT sufficient to make argv inert on this cmd.exe path: libuv still
+    // flattens argv into one CreateProcess command line that cmd.exe re-parses with its own grammar,
+    // and libuv only quotes elements containing space/tab/`"`. A bare metacharacter (e.g. an emulator
+    // name `a&calc`) therefore arrives unquoted and cmd.exe splits on `&`; and even a quoted element
+    // can break out via an embedded `"` (cmd.exe does not honor the CRT `\"` escape — the BatBadBut /
+    // CVE-2024-1874 class), plus `%VAR%` is expanded even inside quotes. Double-quoting does not
+    // reliably neutralize any of these, so we validate each argv element against a strict allowlist
+    // before handing it to cmd.exe. We do NOT use shell:true either: shell:true does not escape argv
+    // (it joins with spaces), which would reintroduce the injection this PR removes.
     //
     // Public (rather than private) solely to act as a substitution seam for unit tests, mirroring
     // CommonUtils.spawnWrapper. Tests stub this to assert that untrusted values are handed over as
     // single, inert argv elements and never reach a shell.
     public static spawnChild(command: string, args: string[] = []): childProcess.ChildProcessWithoutNullStreams {
         if (process.platform === WINDOWS_OS) {
+            // Validate each untrusted argv element against a strict allowlist before it reaches
+            // cmd.exe (see method comment). We deliberately do not validate `command`: it is a
+            // trusted, code-derived SDK path (avdmanager), not user input.
+            args.forEach((arg) => AndroidUtils.assertSafeForCmd(arg));
             const child = childProcess.spawn(process.env.comspec ?? 'cmd.exe', ['/d', '/s', '/c', command, ...args], {
                 shell: false
             });
@@ -898,6 +917,15 @@ export class AndroidUtils {
             });
             child.unref();
             return child;
+        }
+    }
+
+    private static assertSafeForCmd(value: string): void {
+        if (!AndroidUtils.CMD_SAFE_ARG.test(value)) {
+            throw new SfError(
+                `Value cannot be safely passed to cmd.exe: ${JSON.stringify(value)}`,
+                'UnsafeCmdArgument'
+            );
         }
     }
 
--- a/src/common/AndroidUtils.ts
+++ b/src/common/AndroidUtils.ts
@@ -367,15 +367,29 @@ export class AndroidUtils {
         // to generate user friendly display names.
         const resolvedName = emulatorName.replace(/ /gi, '_');
 
-        const createAvdCommand = `${AndroidUtils.getAvdManagerCommand()} create avd -n ${resolvedName} --force -k ${AndroidUtils.systemImagePath(
-            platformAPI,
-            emulatorImage,
+        // Build the command as an argv array so that no value (device name, image path,
+        // device type, abi) is ever re-parsed by a shell. spawnChild runs with shell:false,
+        // so each element below is passed to the OS as a single, inert argument.
+        const avdManagerCommand = AndroidUtils.getAvdManagerCommand();
+        const createAvdArgs = [
+            'create',
+            'avd',
+            '-n',
+            resolvedName,
+            '--force',
+            '-k',
+            AndroidUtils.systemImagePath(platformAPI, emulatorImage, abi),
+            '--device',
+            device,
+            '--abi',
             abi
-        )} --device ${device} --abi ${abi}`;
+        ];
+        // Human-readable form used only for error messages/logging (never executed).
+        const createAvdCommand = `${avdManagerCommand} ${createAvdArgs.join(' ')}`;
 
         return new Promise((resolve, reject) => {
             try {
-                const child = AndroidUtils.spawnChild(createAvdCommand);
+                const child = AndroidUtils.spawnChild(avdManagerCommand, createAvdArgs);
                 if (child) {
                     child.stdin.setDefaultEncoding('utf8');
                     child.stdin.write('no');
@@ -851,6 +865,30 @@ export class AndroidUtils {
         });
     }
 
+    // NOTE: detaching a process in windows seems to detach the streams. Prevent spawn from detaching when
+    // used in Windows OS for special handling of some commands (adb).
+    //
+    // The command is passed as an executable + argv array with shell:false so that no
+    // argument is re-parsed by a shell (structurally removes OS command injection). The
+    // Windows/non-Windows branch exists solely for the `detached` behavior described above.
+    //
+    // Public (rather than private) solely to act as a substitution seam for unit tests, mirroring
+    // CommonUtils.spawnWrapper. Tests stub this to assert that untrusted values are handed over as
+    // single, inert argv elements and never reach a shell.
+    public static spawnChild(command: string, args: string[] = []): childProcess.ChildProcessWithoutNullStreams {
+        if (process.platform === WINDOWS_OS) {
+            const child = childProcess.spawn(command, args, { shell: false });
+            return child;
+        } else {
+            const child = childProcess.spawn(command, args, {
+                shell: false,
+                detached: true
+            });
+            child.unref();
+            return child;
+        }
+    }
+
     private static async getAllCurrentlyUsedAdbPorts(logger?: Logger): Promise<number[]> {
         let ports: number[] = [];
         const command = `${AndroidUtils.getAdbShellCommand()} devices`;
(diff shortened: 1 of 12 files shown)


src/common/CommonUtils.ts
```diff
@@ -312,24 +312,63 @@ export class CommonUtils {
         stdioOptions: StdioOptions = ['ignore', 'pipe', 'ignore']
     ): childProcess.ChildProcess {
         return childProcess.spawn(command, args, {
-            shell: true,
+            // shell:false so that the OS performs no shell parsing of the command or its
+            // arguments. This structurally removes the OS command-injection class: any value
+            // passed as an argv element is treated as an inert argument, never re-interpreted
+            // as shell syntax (`;`, `$()`, backticks, `&&`, `|`, etc.).
+            shell: false,
```
maliroteh-sf, line 328: **Missing `'error'` listener now that this is `shell:false`.** The sibling `spawnCommandAsync` (just below) attaches only a `'close'` handler. Under `shell:true`, a missing/unlaunchable command ran through the shell, which exited non-zero → `'close'` fired → the promise rejected cleanly. Under `shell:false`, a failure to launch the executable itself (ENOENT, EACCES, Windows `.bat` EINVAL) emits an **`'error'`** event instead — and an `'error'` with no listener throws, so the promise never settles.

Every `IOSUtils`/`AppleDevice` sink and `launchUrlInDesktopBrowser` routes through here. Note the new raw-spawn site in `AndroidDevice.ts` *did* get an `'error'` handler, so the failure mode was recognized there but not in the shared primitive.

```ts
prc.on('error', (err) => reject(err));
```

src/common/AndroidUtils.ts
```diff
@@ -851,6 +865,30 @@ export class AndroidUtils {
...
+    // The command is passed as an executable + argv array with shell:false so that no
+    // argument is re-parsed by a shell (structurally removes OS command injection). The
+    // Windows/non-Windows branch exists solely for the `detached` behavior described above.
+    //
+    // Public (rather than private) solely to act as a substitution seam for unit tests, mirroring
+    // CommonUtils.spawnWrapper. Tests stub this to assert that untrusted values are handed over as
+    // single, inert argv elements and never reach a shell.
+    public static spawnChild(command: string, args: string[] = []): childProcess.ChildProcessWithoutNullStreams {
+        if (process.platform === WINDOWS_OS) {
+            const child = childProcess.spawn(command, args, { shell: false });
```
maliroteh-sf: **Windows regression: `avdmanager` is a `.bat`, which `shell:false` can't launch.** `getAvdManagerCommand()` returns an extensionless path; the real file on Windows is `avdmanager.bat`.

- Windows `CreateProcess` auto-appends `.exe` but **not** `.bat`/`.cmd`, so `spawn('…/avdmanager', args, {shell:false})` fails with ENOENT.
- Even with the full `.bat` path, current Node (≥ 18.20.2 / 20.12.2, post CVE-2024-27980) refuses to spawn `.bat`/`.cmd` unless `shell:true` → EINVAL.

Combined with the missing `'error'` handler, on Windows this becomes an unhandled throw. `emulator`/`adb` are `.exe` so they're fine; `sdkmanager` still goes through `executeCommandAsync` (shell) so it's unaffected — which is why this slipped through. Windows CI won't catch it because tests stub `spawnChild`.

Suggest re-enabling `shell:true` on Windows for the `.bat` tools (Node then quotes argv safely for cmd.exe), or invoking via `cmd /c` like the `start` case in `launchUrlInDesktopBrowser`.

OR

Revert back to using `shell:true` EVERYWHERE touched by this PR (if that is simpler)

src/common/IOSUtils.ts
```diff
@@ -170,8 +183,12 @@ export class IOSUtils {
      * @returns The supported iOS device types.
      */
     public static async getSupportedDeviceTypes(logger?: Logger): Promise<string[]> {
-        const command = `${XCRUN_CMD} simctl list devicetypes --json`;
-        const { stdout } = await CommonUtils.executeCommandAsync(command, logger);
+        const { stdout } = await CommonUtils.spawnCommandAsync(
```
maliroteh-sf, line 186: **This newly routes large JSON through `spawnCommandAsync`, which comma-mangles multi-chunk stdout.** `spawnCommandAsync` builds its result with `capturedStdout.join()` — `Array.join()` with no argument uses `,` as the separator. `xcrun simctl list devicetypes --json` emits kilobytes across multiple `'data'` chunks, so commas get injected at each chunk boundary → `JSON.parse(stdout)` throws (or mis-parses).

Short outputs (single chunk, e.g. the UDID from `createNewDevice`) survive, which is why the small-stubbed-stdout tests pass. The bug is latent in the primitive but this PR is what routes JSON output through it. Fix in `CommonUtils.spawnCommandAsync`: use `capturedStdout.join('')` (and the same for the `stderr` resolve and the two error-message joins).

maliroteh-sf (commented):
Review focused on the security refactor's behavioral fallout.

The direction is right: switching the `spawn*` primitives to `shell:false` + argv arrays removes OS command injection structurally, which is stronger than relying on correct quoting everywhere (and is the correct pairing for an API that already takes an argv array). The tests are good — the real-subprocess `$(echo INJECTED)` round-trip is exactly the right way to prove it.

The concern is that flipping `shell:true`→`shell:false` changes runtime behavior for every caller of these primitives, and three of those consequences look like regressions. Details inline.

Lower-severity: `shellQuote`'s Windows branch only doubles `"` and doesn't neutralize cmd.exe metacharacters (`& | < > ^ %`). It's only used on POSIX-only sinks today so it isn't exploitable, but the doc comment implies general Windows safety — worth narrowing.

src/common/CommonUtils.ts
ben-zhang-at-salesforce: Fixed in fe23b3b. Added `prc.on('error', (err) => reject(err))` to `spawnCommandAsync` so a launch failure (ENOENT/EACCES/Windows .bat EINVAL) rejects the promise instead of leaving it unsettled. Test: "spawnCommandAsync rejects when the process emits an error event".

src/common/AndroidUtils.ts
ben-zhang-at-salesforce: Fixed in fe23b3b, but not via the suggested shell:true route. On Windows, spawnChild now launches through cmd.exe: `spawn(comspec, ['/d','/s','/c', command, ...args], { shell: false })` — the same cmd.exe /c pattern already used for `start` in launchUrlInDesktopBrowser.

I did not re-enable shell:true because shell:true does NOT escape argv — Node joins command + args with spaces and sets windowsVerbatimArguments, so an emulator name like `a & calc` would inject. That would reintroduce exactly the class of bug this PR removes. With shell:false + cmd.exe /c, Node quotes each argv element and cmd.exe honors those double quotes for the RCE metacharacters (& | ; < >), so untrusted values stay inert while cmd.exe still resolves avdmanager -> avdmanager.bat via PATHEXT. Test: 'spawnChild passes a malicious argv element inertly'.

src/common/IOSUtils.ts
ben-zhang-at-salesforce: Fixed in fe23b3b. spawnCommandAsync now joins with `''` instead of the default `','` for the two resolve paths and the two error-message joins, so multi-chunk stdout (e.g. xcrun simctl list devicetypes --json across several 'data' events) is reassembled verbatim and JSON.parse succeeds. Test: 'spawnCommandAsync reassembles multi-chunk stdout verbatim (no comma separator)' emits two chunks and asserts JSON.parse does not throw.

src/common/AndroidUtils.ts
maliroteh-sf: Hmm... claude is saying that the issue is still not fixed. Here are its findings:

The `cmd.exe /d /s /c` route is the correct shape and it does solve the launch problem (PATHEXT resolves `avdmanager → avdmanager.bat`). But two things keep this from matching the "inert" claim on Windows:

1. Embedded-double-quote breakout (BatBadBut / CVE-2024-1874 class). Node quotes argv with CRT rules (escaping `"` as `\"`), but cmd.exe doesn't honor `\"` — it sees the quote as closing quote-context.
Concretely, for an emulator name `a"&calc`:
Node emits:   `"a\"&calc"`
cmd.exe sees: " → quote on, a\ literal, " → quote OFF, & → separator, calc runs
So a value with a `"` followed by a metacharacter escapes. Reachable via `resolvedName` (the emulator name is only space→underscore-normalized; `" & %` survive) and `device`. Lower likelihood than the original bug and the companion whitelist in #302 catches bad device types — but the emulator name isn't whitelisted and  `createNewVirtualDevice` is callable directly. Suggest rejecting/sanitizing argv elements containing `"` before the cmd.exe invocation (and consider `%`, which cmd expands even inside quotes).

2. The new test doesn't exercise the cmd.exe path it's meant to protect. `spawnChild passes a malicious argv element inertly` uses `'$(echo INJECTED)'` — POSIX command-substitution syntax that cmd.exe doesn't interpret, so it arrives literally on Windows no matter what. The test would still pass even if `a"&calc` broke out. To actually prove cmd.exe safety, use a cmd-native payload (e.g. `a & echo INJECTED`, or an embedded-quote case). Windows CI does run this (`run-tests.yml` has `windows-latest`), so a cmd-native payload would give it real teeth.

maliroteh-sf: And it is suggesting to fix it as below:

**The core idea**

The residual risk is only on the **Windows** `cmd.exe /c` path. The clean fix is: **when we route through cmd.exe, sanitize the argv elements against cmd's grammar before handing them over.** The POSIX branch stays untouched (already inert via `execve`).

---
**Change 1 — add a Windows cmd.exe argv guard + use it in `spawnChild`**
**File**: `src/common/AndroidUtils.ts` — replace the `spawnChild` method
```
      // Characters that are safe to pass to cmd.exe as (Node-quoted) argv elements.
      // Allowlist rather than denylist: the values that reach the Windows cmd.exe path
      // are structured (an already-normalized emulator name, a device id, a
      // `system-images;api;image;abi` path, an abi). Anything outside this set is
      // rejected rather than risk cmd.exe re-interpretation (`& | < > ^ ( )`,
      // `%VAR%` expansion, `!` delayed expansion), none of which double-quoting
      // reliably neutralizes.
      private static readonly CMD_SAFE_ARG = /^[A-Za-z0-9 ._:;@\\/=+-]+$/;

      private static assertSafeForCmd(value: string): void {
          if (!AndroidUtils.CMD_SAFE_ARG.test(value)) {
              throw new SfError(
                  `Value cannot be safely passed to cmd.exe: ${JSON.stringify(value)}`,
                  'UnsafeCmdArgument'
              );
          }
      }

      public static spawnChild(command: string, args: string[] = []): childProcess.ChildProcessWithoutNullStreams {
          if (process.platform === WINDOWS_OS) {
              // shell:false still flattens argv into one CreateProcess command line that cmd.exe
              // re-parses with its own grammar, and libuv only quotes args containing space/tab/".
              // So a bare metacharacter (e.g. an emulator name `a&calc`) would break out. Validate
              // each untrusted element against a strict allowlist before handing it to cmd.exe.
              args.forEach((arg) => AndroidUtils.assertSafeForCmd(arg));
              const child = childProcess.spawn(process.env.comspec ?? 'cmd.exe', ['/d', '/s', '/c', command, ...args], {
                  shell: false
              });
              return child;
          } else {
              const child = childProcess.spawn(command, args, {
                  shell: false,
                  detached: true
              });
              child.unref();
              return child;
          }
      }
```
Notes:
- `SfError` is already imported in this file (line 10) — no new import.
- I deliberately don't validate `command` (the `avdmanager` path); it's a trusted, code-derived SDK path, not user input.
- The allowlist includes `;` and `=` because the image path is `system-images;android-33;google_apis;x86_64` and launch-style `name=value` args exist elsewhere; it includes space because `resolvedName` only strips spaces for AVDs but other callers might not. It excludes every cmd metacharacter.

---
**Change 2 — apply the same guard to the other cmd.exe sink (`launchUrlInDesktopBrowser`)**

This sink has the identical flaw, and for URLs it's also a functional bug: an ordinary `http://h/?a=1&b=2` has no space, so libuv leaves it unquoted and cmd splits on `&`. A URL can't go through a plain allowlist (it legitimately contains `?`, `&`, `#`, `%`), so here the right move is force-quote and reject only what quoting can't save (`"`, `%`, CR/LF).

**File**: `src/common/CommonUtils.ts` — in `launchUrlInDesktopBrowser`, the Windows branch.
```
          } else if (process.platform === 'win32') {
              // `start` is a cmd.exe builtin, so it must run via cmd.exe. libuv only quotes argv
              // elements containing space/tab/", and cmd.exe treats & | < > ^ ( ) as metacharacters
              // and expands %VAR% even inside quotes. A normal URL (`?a=1&b=2`) has no space, so it
              // would be split on `&`. Wrap the URL in double quotes ourselves, and refuse the two
              // things quoting cannot make safe: an embedded " (closes cmd's quote context) and
              // %VAR% expansion (and CR/LF).
              if (/["%\r\n]/.test(url)) {
                  throw new SfError(`URL cannot be safely opened on Windows: ${JSON.stringify(url)}`, 'UnsafeUrl');
              }
              command = 'cmd';
              args = ['/c', 'start', '', `"${url}"`];
          } else {
```

---
**Change 3 — fix the test so it actually exercises the Windows path**

The current test uses `$(echo INJECTED)` — POSIX substitution syntax cmd.exe ignores, so it passes on Windows even if a breakout exists. Add a cmd-native assertion that the guard rejects a bare metacharacter.

**File**: `test/unit/common/AndroidUtils.test.ts`
```
      it('spawnChild rejects a cmd.exe metacharacter argv element on Windows', function () {
          if (process.platform !== 'win32') {
              this.skip(); // guard is Windows-only; POSIX passes these to execve inertly by design
          }
          // `a&calc` has no space/tab/" so libuv would leave it unquoted and cmd.exe would split on &.
          expect(() => AndroidUtils.spawnChild('avdmanager', ['create', 'avd', '-n', 'a&calc'])).to.throw(
              /cannot be safely passed to cmd\.exe/
          );
      });
```

Since CI has a `windows-latest` job (`run-tests.yml`), this gives the protection real teeth there while being a no-op on the Ubuntu/macOS runners.

ben-zhang-at-salesforce: Fixed in 4a2037e. Both residual issues you raised are addressed, and the direction you outlined is what I implemented.

**1. cmd.exe metacharacter / embedded-quote breakout (BatBadBut / CVE-2024-1874 class).** Confirmed: shell:false is not sufficient on the cmd.exe /c path because libuv only quotes argv elements containing space/tab/`"`, so a bare `a&calc` arrives unquoted and cmd splits on `&`, and an embedded `"` (`a"&calc`) escapes cmd's quote context regardless. `spawnChild` now validates each argv element against a strict allowlist (`/^[A-Za-z0-9 ._:;@\\/=+-]+$/`) before handing it to cmd.exe and throws `UnsafeCmdArgument` otherwise. The `command` path (`avdmanager`) is not validated — it's a trusted, code-derived SDK path. `%`, `!`, and every cmd metacharacter are excluded; `;` and `=` are allowed for the `system-images;api;image;abi` path.

**2. Test didn't exercise the Windows path.** Confirmed — `$(echo INJECTED)` is POSIX substitution cmd.exe ignores, so it would pass on Windows even with a breakout. That test is now scoped to non-Windows, and a new Windows-only test asserts the guard rejects cmd-native payloads (`a&calc` and the embedded-quote `a"&calc`). Since `run-tests.yml` runs `windows-latest`, this has real teeth there.

**Also fixed the identical flaw in the other cmd.exe sink** you flagged, `launchUrlInDesktopBrowser`: a normal URL like `?a=1&b=2` has no space so libuv left it unquoted and cmd split on `&` (a functional bug too). The URL is now force-quoted and the Windows branch rejects what quoting can't neutralize (`"`, `%`, CR/LF) with `UnsafeUrl`. Added a Windows-only rejection test for that as well.

Full suite: 211 passing, 2 pending (the Windows-only tests, skipped on macOS/Ubuntu), lint clean.

maliroteh-sf (approved):
Looks good to me now. Great job 👍

PR #221, opened 2026-08-14
fix: harden residual command-injection sinks + validate device names (@W-23665968@)

haifeng-li-at-salesforce:
## What & why

Completes the OS command-injection hardening for `lwc-dev-mobile-core` that **W-23665289** (commit `d1482b2`, #220) began. That commit converted the child-process spawn primitives to `shell:false` + argv arrays and shell-quoted the genuine pipe sinks. This PR closes the two gaps it left:

### 1. `CommonUtils.extractZIPArchive` — the last unquoted shell-string sink
It built a shell command by interpolating untrusted paths (`unzip -o -qq ${archive} -d ${outDir}` on POSIX; a cmd/PowerShell-unsafe string on Windows) and ran it through `executeCommandAsync` (a shell). Reworked to run via `spawnCommandAsync` (`shell:false`):
- **POSIX** — `unzip` with each path as an inert argv element.
- **Windows** — `Expand-Archive` through `powershell` (a real executable, no cmd.exe), with each path wrapped in a PowerShell single-quoted literal (embedded `'` doubled), inside which PowerShell performs no expansion or command substitution. Adds `CommonUtils.powerShellSingleQuote` for that.

### 2. Device-name / identifier input validation (report recommendation #3)
Never implemented before. Adds `CommonUtils.assertSafeDeviceName` / `assertSafeDeviceIdentifier` — strict allowlists that throw `SfError` on shell metacharacters — called at the top of `AndroidUtils.createNewVirtualDevice` and `IOSUtils.createNewDevice`, so malformed input fails fast before any command is built. Defense-in-depth on top of the `shell:false` execution.

## Scope notes
- 3 of the 12 sinks the report lists (`run.ts`, `configureLintingToolsCommand.ts`, `orgUtils.ts`) **do not exist in this repo** — they live in sibling packages.
- The `AppleDevice` / `MacNetworkUtils` pipe sinks were already `shellQuote`-guarded by W-23665289 and are unchanged.

## Testing
TDD throughout (tests written failing-first). Added coverage for the validators (accept/reject), fail-fast rejection in both device-creation methods, and `extractZIPArchive` argv / PowerShell-quoting behavior.

- `yarn compile` ✅
- `yarn lint` ✅ (0 errors; 2 pre-existing warnings in untouched `AppleDeviceManager.ts`)
- `yarn test` ✅ 222 passing, coverage above all thresholds

GUS: https://gus.lightning.force.com/lightning/r/a07EE00002gJcjXYAS/view

### DIFF
--- a/src/common/CommonUtils.ts
+++ b/src/common/CommonUtils.ts
@@ -24,6 +24,17 @@ const messages = Messages.loadMessages('@salesforce/lwc-dev-mobile-core', 'commo
 export class CommonUtils {
     public static DEFAULT_LWC_SERVER_PORT = '3333';
 
+    // Allowlist for user-facing device/emulator names: letters, digits, space, and the few
+    // punctuation characters that legitimately appear (`.`, `_`, `-`). Defense-in-depth on top of
+    // the shell:false argv execution — it fails fast on unexpected input and excludes every shell
+    // metacharacter (`; & | $ ( ) \` < > * ? ! # ' " \\` etc.) so a name can never be repurposed
+    // even if a future caller reintroduces a shell sink.
+    private static readonly SAFE_DEVICE_NAME = /^[A-Za-z0-9 ._-]+$/;
+
+    // Allowlist for structured device identifiers (deviceType, runtime, abi). Same as a name but
+    // with no spaces, since these are token-like values (e.g. `iPhone-8`, `iOS-17-2`, `x86_64`).
+    private static readonly SAFE_DEVICE_IDENTIFIER = /^[A-Za-z0-9._-]+$/;
+
     /**
      * Converts a path to UNIX style path.
      *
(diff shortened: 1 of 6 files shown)


src/common/CommonUtils.ts
```diff
@@ -356,6 +367,51 @@ export class CommonUtils {
...
+     * so this is not the sole barrier — but it rejects malformed input up front and guarantees a
+     * name can never carry a shell metacharacter into any current or future sink.
+     *
+     * @param name The device/emulator name to validate.
+     */
+    public static assertSafeDeviceName(name: string): void {
+        if (!CommonUtils.SAFE_DEVICE_NAME.test(name)) {
+            throw new SfError(`Device name cannot be safely used: ${JSON.stringify(name)}`, 'UnsafeDeviceName');
+        }
+    }
```
maliroteh-sf, line 397: This may be too strict. It is rather common to have names like `Pixel 6 API 33` but I've also seen `Pixel 6 (API 33)` which now is not going to be allowed because `(` `)` are not allowed anymore. But this may be ok as long as the errors that we return are more descriptive. I probably should have flagged this in the PR that @ben-zhang-at-salesforce authored too.

Rather than the generic `... cannot be safely used` we should probably say something like:

`assertSafeDeviceName`:
```
throw new SfError(
    `Invalid value ${JSON.stringify(name)}: only letters, numbers, spaces, and . _ - are allowed.`,
    'UnsafeDeviceName'
);
```

`assertSafeDeviceIdentifier`:
```
throw new SfError(
    `Invalid value ${JSON.stringify(identifier)}: only letters, numbers, and . _ - are allowed (no spaces).`,
    'UnsafeDeviceIdentifier'
);
```

`assertSafeForCmd`:
```
throw new SfError(
    `Invalid value ${JSON.stringify(value)}: only letters, numbers, spaces, and . _ - : ; @ \\ / = + are allowed when launching this command on Windows.`,
    'UnsafeCmdArgument'
);
```

`launchUrlInDesktopBrowser`:
```
  throw new SfError(
      `Invalid value ${JSON.stringify(url)}: URLs containing a double quote, a percent sign, or a newline cannot be opened safely on Windows.`,
      'UnsafeUrl'
  );
```

haifeng-li-at-salesforce: Good catch — updated all four error messages (`assertSafeDeviceName`, `assertSafeDeviceIdentifier`, `assertSafeForCmd`, `launchUrlInDesktopBrowser`) to spell out the allowed character set / rejected characters instead of the generic "cannot be safely used" wording, using the text you suggested. Allowlists themselves are unchanged (still no parens) — just the messaging, so users get an actionable reason for the rejection. Pushed in 62df9e5.

maliroteh-sf (approved):

